PR #9, opened 2022-09-08
Github action to auto deploy to PyPI

aryanagarwal9:

feature

marqo needs to be manually deployed to PyPI.

The workflow is triggered every time a change is pushed/merged into the 'mainline' branch, and the package is built and then published to PyPI.

no

* **Other information**:
a token named "PYPI_API_TOKEN" needs to be present in GitHub secrets for authentication purposes.
the workflow will fail if setup.py is not updated with the new version number

### DIFF
--- a/.github/workflows/publish-to-PyPI.yml
+++ b/.github/workflows/publish-to-PyPI.yml
@@ -1,0 +1,37 @@
+name: Publish to PyPI
+
+on:
+  push:
+    branches:
+      - mainline
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+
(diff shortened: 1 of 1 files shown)


pandu-k:
This looks really cool! Just have one question:

Does this handle the PyPi push failing? For example, if we push to mainline, but neglect to bump the version number, PyPi rejects the push. What happens then?

It could be nice if it accepted the push to mainline, gracefully handling the pypi upload failure, but sent our team an alert (just slack is fine) about the failure. This might already happen.

pandu-k (commented):
looks good!

PR #42, opened 2022-12-13
Cloud defaults

pandu-k:

Feature - specific set of defaults for creating indices on the cloud, refactor

Cloud indices are set up with the same defaults as non cloud indices

Cloud indices are set up with cloud-specific settings (specifically, number of shards = 2)

No

* **Other information**:

__Manual testing__ 
1. Confirmed that an index that with a specified model (multilingual), creates and index with 2 shards.
2. Confirmed that an index with default params (`mq.create_index(index_name=self.index_name_1)`) creates an index with 2 shards:
image

### DIFF
--- a/src/marqo/defaults.py
+++ b/src/marqo/defaults.py
@@ -1,0 +1,22 @@
+
+def get_cloud_default_index_settings():
+    """
+    Diverges from default in:
+        - number_of_shards
+    """
+    return {
+        "index_defaults": {
+            "treat_urls_and_pointers_as_images": False,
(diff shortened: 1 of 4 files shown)


src/marqo/defaults.py
```diff
@@ -0,0 +1,22 @@
+
+def get_cloud_default_index_settings():
```
Jeadie, line 2: It could be good, eventually, to move this to a mountable config.

Jeadie (approved):

PR #60, opened 2023-02-20
Image download headers

pandu-k:

Feature

No way to propagate images that require auth headers to access

Propagate image auth headers via `image_download_headers` in search() and add_documents()

No

* **Other information**:
- Manually tested add_docs and search workflows

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -234,7 +234,8 @@ class Index:
         processes: int = None,
         device: str = None,
         non_tensor_fields: List[str] = None,
-        use_existing_tensors: bool = False
+        use_existing_tensors: bool = False,
+        image_download_headers: dict = None
     ) -> Union[Dict[str, Any], List[Dict[str, Any]]]:
         """Add documents to this index. Does a partial update on existing documents,
         based on their ID. Adds unseen documents to the index.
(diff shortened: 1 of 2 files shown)


tomhamer (approved):

src/marqo/index.py
```diff
@@ -116,7 +116,7 @@ def refresh(self):
     def search(self, q: Union[str, dict], searchable_attributes: Optional[List[str]] = None,
                limit: int = 10, offset: int = 0, search_method: Union[SearchMethods.TENSOR, str] = SearchMethods.TENSOR,
                highlights=None, device: Optional[str] = None, filter_string: str = None,
-               show_highlights=True, reranker=None,
+               show_highlights=True, reranker=None, image_download_headers: Optional[Dict] = None,
```
wanliAlex, line 119: do we need to specify the type of the keys and values, like Dict[str, str]?

wanliAlex (changes requested):
do we need to add tests?

src/marqo/index.py
pandu-k: I think we leave it quite flexible - because the header details can be different depending on customer use case

wanliAlex (approved):

PR #63, opened 2023-02-22
Reducing logging output

pandu-k:

Logging update

Too much info-level logging

Logs are only output for client-side batching, at a rate of one log line per batch. The rest of ingest and search logging has been moved to DEBUG

No

* **Other information**:
See comments below for log outputs at the INFO and DEBUG levels.

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -183,11 +183,12 @@ class Index:
         end_time_client_request = timer()
         total_client_request_time = end_time_client_request - start_time_client_request
         
-        search_time_log = f"search ({search_method.lower()}): took {(total_client_request_time):.3f}s to send query and received {num_results} results from Marqo (roundtrip)."
+        search_time_log = (f"search ({search_method.lower()}): took {(total_client_request_time):.3f}s to send query "
+                           f"and received {num_results} results from Marqo (roundtrip).")
         if 'processingTimeMs' in res:
             search_time_log += f" Marqo itself took {(res['processingTimeMs'] * 0.001):.3f}s to execute the search."
         
-        mq_logger.info(search_time_log)
+        mq_logger.debug(search_time_log)
         return res
 
     def get_document(self, document_id: str, expose_facets=None) -> Dict[str, Any]:
@@ -354,7 +355,8 @@ class Index:
         )
         end_time_client_process = timer()
         total_client_process_time = end_time_client_process - start_time_client_process
-        mq_logger.info(f"add_documents pre-processing: took {(total_client_process_time):.3f}s for {num_docs} docs, for an average of {(total_client_process_time / num_docs):.3f}s per doc.")
+        mq_logger.debug(f"add_documents pre-processing: took {(total_client_process_time):.3f}s for {num_docs} docs, "
+                       f"for an average of {(total_client_process_time / num_docs):.3f}s per doc.")
 
         if client_batch_size is not None:
             if client_batch_size <= 0:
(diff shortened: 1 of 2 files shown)


pandu-k:
### client_batch_size=50, server_batch_size=10, No server MP


**INFO output**


image

#### Debug output

image

pandu-k:
### client_batch_size=50, server_batch_size=10, processes=2

**INFO output:**
image

#### Debug output

image

pandu-k:
### client_batch_size=50, processes=2, No server_batch_size
#### INFO output:

image

#### DEBUG output:
image

pandu-k:
### client_batch_size=50, No server MP, No server_batch_size
#### INFO output: 
image

#### DEBUG output:
image

pandu-k:
### NO client_batch_size, No server MP, No server_batch_size
#### INFO output:
n/a (None)
#### DEBUG output:
image

pandu-k:
### Search output
#### INFO output:
n/a (None)
#### DEBUG output:
image

pandu-k:
Unit tests pass

src/marqo/index.py
```diff
@@ -499,35 +501,37 @@ def verbosely_add_docs(i, docs):
                                  f'Allowed update_methods: ["replace", "update"] ')
             total_batch_time = timer() - t0
             num_docs = len(docs)
-            mq_logger.info(
```
vicilliar: moving this line into the `if isinstance(res, list)` means that it will only output if there is server batching, but the line was intended to be a report of a client batch.

vicilliar: also, maybe this is the only message we want to keep as "info", since it's a summary of a client batch, regardless of whether there's server batching or not.

src/marqo/index.py
```diff
@@ -499,35 +501,37 @@ def verbosely_add_docs(i, docs):
...
-                        mq_logger.info(f"       marqo server batch {batch}: "
+                        mq_logger.debug(f"       marqo server batch {batch}: "
                                         f"processed {server_batch_result_count} docs in {(res[batch]['processingTimeMs'] / 1000):.3f}s, "
                                         f"for an average of {(res[batch]['processingTimeMs'] / (1000 * server_batch_result_count)):.3f}s per doc.")
             else:
                 # no Server Batching
                 if 'processingTimeMs' in res:       # Only outputs log if response is non-empty
-                    mq_logger.info(f"   add_documents batch {i} Marqo processing: took {(res['processingTimeMs'] / 1000):.3f}s for Marqo to process & index {num_docs} "
-                                    f"docs (server unbatched), for an average of {(res['processingTimeMs'] / (1000 * num_docs)):.3f}s per doc.")
+                    mq_logger.info(
```
vicilliar: if we only want 1 message to remain "info" per batch, do we want it to be this (processing time as reported by Marqo), or the one at line 502 (measured roundtrip time)?

src/marqo/marqo_logging.py
```diff
@@ -6,6 +6,8 @@
 formatter = logging.Formatter(
     "{asctime} logger:'marqo' {levelname} {message}", style='{')
 ch = logging.StreamHandler()
-ch.setLevel(logging.INFO)
+
+# the ch stream handler won't accept any logs from a higher level than the one set here:
+ch.setLevel(logging.DEBUG)
```
vicilliar: does this override the above line `mq_logger.setLevel(logging.INFO)`

pandu-k: Basically they are two decoupled objects. It seems that the handler won't accept any logs at a higher level than its level. The logger won't accept any incoming logs higher than its log level. 

Before, because ch.Level=INFO, you could never output DEBUG statements coming from the logger to the console

src/marqo/index.py
pandu-k: Makes sense, but this is my thinking:

- for the server batching it doesn't make sense to have the Marqo execution time (since it is more informative to have this information per server batch, which we can get from DEBUG), so roundtrip time is enough. 
- For the client batching, we want to have the roundtrip and execution time in the same line. 

That's why I split these two info statements out

src/marqo/index.py
pandu-k: See my comment here: https://github.com/marqo-ai/py-marqo/pull/63/files/9d0b1c45c67df99089784a5cbb6ec114be7c90af#r1115317936

src/marqo/marqo_logging.py
vicilliar: ok, that checks out!

src/marqo/index.py
vicilliar: ok makes sense!

vicilliar (approved):

PR #91, opened 2023-05-22
Bulk search now supports context and score modifiers

Jeadie:
Client side support for [PR #469](https://github.com/marqo-ai/marqo/pull/469). Enabled the following in bulk search: 
1. Context vectors
2. Score modifiers

### DIFF
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -12,22 +12,25 @@ from marqo.errors import (
 
 HTTP_OPERATIONS = Literal["delete", "get", "post", "put"]
 ALLOWED_OPERATIONS: Tuple[HTTP_OPERATIONS, ...] = get_args(HTTP_OPERATIONS)
+session = requests.Session()
 
-OPERATION_MAPPING = {'delete': lambda s: s.delete, 'get': lambda s: s.get,
-                     'post': lambda s: s.post, 'put': lambda s: s.put}
-
+OPERATION_MAPPING = {
+    'delete': session.delete,
+    'get': session.get,
+    'post': session.post,
+    'put': session.put
+}
 
 class HttpRequests:
-    def __init__(self, config: Config, s: requests.session = requests.Session()) -> None:
+    def __init__(self, config: Config) -> None:
         self.config = config
-        self.session = s
         self.headers = {'x-api-key': config.api_key} if config.api_key else {}
 
     def _operation(self, method: HTTP_OPERATIONS) -> Callable:
         if method not in ALLOWED_OPERATIONS:
             raise ValueError("{} not an allowed operation {}".format(method, ALLOWED_OPERATIONS))
 
-        return OPERATION_MAPPING[method](self.session)
+        return OPERATION_MAPPING[method]
 
     def send_request(
         self,
@@ -38,7 +41,7 @@ class HttpRequests:
     ) -> Any:
         req_headers = copy.deepcopy(self.headers)
 
-        if content_type:
+        if content_type is not None and content_type
             req_headers['Content-Type'] = content_type
 
         if not isinstance(body, (bytes, str)) and body is not None:
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -1,6 +1,6 @@
 import copy
 import json
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import get_args, Any, Callable, Dict, Literal, List, Optional, Tuple, Union
 import requests
 from json.decoder import JSONDecodeError
 from marqo.config import Config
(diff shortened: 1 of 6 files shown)


src/marqo/_httprequests.py
```diff
@@ -1,6 +1,6 @@
 import copy
```
Jeadie, line 1: I had to simplify a bunch of this to allow for better mocking.

src/marqo/models.py
```diff
@@ -26,10 +26,7 @@ class SearchBody(BaseMarqoModel):
 
 class BulkSearchBody(SearchBody):
     index: str
-    # Attributes that are not supported in bulk search
```
Jeadie, line 29: This is the only real change

src/marqo/_httprequests.py
```diff
@@ -10,90 +10,71 @@
...
     ) -> Any:
-        to_verify = True
-
-        if http_method not in ALLOWED_OPERATIONS:
-            raise ValueError("{} not an allowed operation {}".format(http_method, ALLOWED_OPERATIONS))
-
         req_headers = copy.deepcopy(self.headers)
 
-        if content_type is not None and content_type:
+        if content_type:
```
pandu-k: I prefer `if content_type is not None and content_type:` for readability reasons

src/marqo/_httprequests.py
```diff
@@ -10,90 +10,71 @@
...
-                    data=json.dumps(body) if body else None,
-                    verify=to_verify
-                )
+            response = self._operation(http_operation)(
+                url=f"{self.config.url}/{path}",
+                timeout=self.config.timeout,
+                headers=req_headers,
+                data=body,
+                verify=True
+            )
```
pandu-k, line 57: Good simplification, the only issue is that, in the existing code, if the body is Falsey, then body is set to None. For example if `body=""`  it won't get propagated into the request.

src/marqo/_httprequests.py
```diff
@@ -10,90 +10,71 @@
     BackendTimeoutError
 )
 
-s = requests.Session()
```
pandu-k, line 13: What's the impact of this being moved from a global to the object level?

tests/v0_tests/test_bulk_search.py
```diff
@@ -111,61 +197,37 @@ def test_search_highlights(self):
...
-        ])
-
-        if self.IS_MULTI_INSTANCE:
-            self.warm_request(self.client.bulk_search, [{
-                "index": self.index_name_1,
-                "q": "Examples of leadership"
-            }])
-
+        }]}, warmup_query="Examples of leadership")
+    def test_select_lexical(self, index_1_docs):
```
pandu-k, line 230: This search isn't even lexical (acknowledged that this naming issue was introduced a while ago)

tests/marqo_test.py
pandu-k, line 1: Could we also have an end-to-end test of bulk search with `context` and `scoreModifiers`

pandu-k: include `warm_request()` calls. This is because py-marqo unit tests are also used as integration tests for cloud.

src/marqo/_httprequests.py
VitusAcabado: There shouldn't be any impact if the `Session()` object used for sending HTTP requests is the same instantiated object in the `HttpRequests` class. As long as the `HttpRequests()` object initally created is the same each time we send a request via the `self_request` operation, it is effectively using the same `Session` object.

If a new `Session` object is created each time, we are basically making a new connection to the upstream server, **which we do not want** (as to avoid possible 503s due to TCP RST packet conflict)

pandu-k: Thanks @VitusAcabado . Looking into this, it seems that it is best to keep this global. Especially when we consider the following use case:
```
mq = marqo.Client()
mq.index('my-index').add_documents([{"Title": "Hello world"}])
mq.index('my-index').search('hello')
```
Line 2,3 both instantiate an Index Object, which in turn instantiates an HTTPRequest so there is no optimisation to be had. There would be a new HTTPRequest, and therefore session object, per line. See here for reference: https://github.com/marqo-ai/py-marqo/blob/mainline/src/marqo/client.py#L107

src/marqo/_httprequests.py
Jeadie: Should we be sending `body=""`

tests/marqo_test.py
Jeadie: What why are we using a client unit tests for a cloud integration test?

src/marqo/_httprequests.py
Jeadie: This is a good spot. I think `def __init__(self, config: Config, s: requests.session = session) -> None:` will solve the issue (i.e. we can pass in session for testing, and use the global variable `session` more regularly).

tests/v0_tests/test_bulk_search.py
Jeadie: It is testing lexical? 
```python
        # try it with lexical search:
        #    can't find the above with synonym
        assert len(self.client.index(self.index_name_1).search(
            "Examples of leadership", search_method=marqo.SearchMethods.LEXICAL)["hits"]) == 0
        #    but can look for a word
        assert self.client.index(self.index_name_1).search(
            '"captain"')["hits"][0]["_id"] == "123456"
```
Essentially checking `search_method=marqo.SearchMethods.LEXICAL` means synonym queries will have zero hits.

tests/marqo_test.py
pandu-k: @Jeadie this was a decision made by the cloud team, as it is the most straightforward way to do Marqo tests

src/marqo/_httprequests.py
pandu-k: Have you thought about the impact?

pandu-k: Also what about `body=False`

Jeadie: I don't think this is a huge problem given the typing given and the fact it's an internal usage that we handle

tests/v0_tests/test_custom_vector_search.py
```diff
@@ -75,27 +79,67 @@ def test_custom_search_results(self):
...
         except MarqoWebError as e:
-            assert "The provided vectors are not in the same dimension of the index" in str(e)
+            assert "This is not supported as the context only works when the query is a dictionary." in str(e)
+        finally:
+
+            ## Ensure other tests are not affected
+            self.query = {"What are the best pets": 1}
+
+class TestCustomBulkVectorSearch(TestCustomVectorSearch):
+
```
pandu-k, line 135: This doesn't seem to be fully implemented

Jeadie: What do you mean?

wanliAlex (approved):

PR #99, opened 2023-06-07
Fix warm up tests in search test suites

Jeadie:
## Changeset
  - Fix search tests when `IS_MULTI_INSTANCE`
  - Change in vector dimension is required for correct testing.

## Testing
 - Explicitly overrode [cls.IS_MULTI_INSTANCE](https://github.com/marqo-ai/py-marqo/blob/mainline/tests/marqo_test.py#L119) to both True and False.

### DIFF
--- a/tests/v0_tests/test_custom_vector_search.py
+++ b/tests/v0_tests/test_custom_vector_search.py
@@ -29,6 +29,7 @@ class TestCustomVectorSearch(MarqoTestCase):
                 }
             ]
         )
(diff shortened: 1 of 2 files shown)


pandu-k (commented):
Looks good. Can you please add what type of testing has been done for this change [here](https://github.com/marqo-ai/py-marqo/pull/99#issue-1745167765)?

pandu-k (approved):

PR #100, opened 2023-06-07
Changes to tests to prevent leftover test indexes

pandu-k:
Changes added to prevent dangling indexes created during testing.

Tests update

Some indexes aren't torn down properly after testing is complete

All indexes are torn down properly after testing is complete. Testing done against the cloud confirms this.

no

* **Other information**:
- tests pass against local Marqo instance, and confirmed that no indexes remained
- tested on the cloud, confirmed that no indexes remained

### DIFF
--- a/tests/marqo_test.py
+++ b/tests/marqo_test.py
@@ -11,6 +11,7 @@ from pydantic import BaseModel
 from typing import Any, Callable, Dict, List, Optional, Union
 from unittest import mock, TestCase
 
+import marqo
 from marqo.utils import construct_authorized_url
 from marqo._httprequests import HTTP_OPERATIONS
 from marqo.version import __marqo_version__ as py_marqo_support_version
@@ -114,6 +115,7 @@ class MarqoTestCase(TestCase):
 
         cls.client_settings = local_marqo_settings
         cls.authorized_url = cls.client_settings["url"]
+        cls.generic_test_index_name = 'test-index'
 
         # class property to indicate if test is being run on multi
         cls.IS_MULTI_INSTANCE = (True if os.environ.get("IS_MULTI_INSTANCE", False) in ["True", "TRUE", "true", True] else False)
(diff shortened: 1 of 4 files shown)


tests/marqo_test.py
```diff
@@ -124,7 +126,18 @@ def setUpClass(cls) -> None:
             print(f"MARQO SERVER VERSION -> {marqo_server_version}")
             print(f"PY-MARQO SUPPORTED VERSION -> {py_marqo_support_version}")
 
-
+    @classmethod
+    def tearDownClass(cls) -> None:
+        """Delete commonly used test indexes after all tests are run
+        """
+        client = marqo.Client(**cls.client_settings)
+        commonly_used_ix_name = 'my-test-index-1'
```
Jeadie, line 135: I feel like we should have a single place where we define the indexes a class will be using. MarqoTestCase can have some generic ones, and subclasses can override this by "appending" more.

tests/marqo_test.py
```diff
@@ -124,7 +126,18 @@ def setUpClass(cls) -> None:
...
+        indexes_to_tear_down = [cls.generic_test_index_name, commonly_used_ix_name]
+        for ix_name in indexes_to_tear_down:
+            try:
+                client.delete_index(ix_name)
+            except marqo.errors.MarqoApiError:
+                pass
```
Jeadie: nit: Generally log errrors, even if debug/info/warning

tests/v0_tests/test_image_chunking.py
```diff
@@ -16,7 +16,7 @@ class TestImageChunking(MarqoTestCase):
     def setUp(self) -> None:
         client_0 = Client(**self.client_settings)
         
-        self.index_name = 'image-chunk-test'
```
Jeadie, line 19: As below, I assume these don't change the tests.

Jeadie: Following on from above comment, if we have class defined indexes used. we could then just reference `self.index_name  = TestImageChunking.generic_test_index_name` or `self.index_name  = TestImageChunking.generic_test_index_name[-1]`

tests/v0_tests/test_demos.py
```diff
@@ -189,7 +189,7 @@ def test_readme_example_weighted_query(self):
     def test_readme_example_multimodal_combination_query(self):
         import marqo
         mq = marqo.Client(**self.client_settings)
-        settings = {"treat_urls_and_pointers_as_images": True, "model": "ViT-L/14"}
+        settings = {"treat_urls_and_pointers_as_images": True, "model": "ViT-B/32"}
```
Jeadie, line 192: I assume this change is to just improve throughput?

Jeadie (approved):

PR #103, opened 2023-06-19
Removing  default device in client (related to Marqo CUDA device PR)

vicilliar:

Feature

Currently, if no device is set when calling add docs, search, or bulk search functions, a default is selected.

- All default devices were removed
- Devices were removed from client config
- When no device is passed in client function, no device parameter is passed to marqo. Marqo itself will do all default calculation. Related PR here: https://github.com/marqo-ai/marqo/pull/508

Yes. Previously, setting no device would usually default to the client sending `cpu` as device. Now it will send no device.

### DIFF
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -19,7 +19,7 @@ class Config:
         Parameters
         ----------
         url:
-            The url to the Marqo instance (ex: http://???)
+            The url to the Marqo instance (ex: http://localhost:8882)
         """
         self.cluster_is_remote = False
         self.cluster_is_s2search = False
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -13,25 +13,19 @@ class Config:
         self,
         url: str,
         timeout: Optional[int] = None,
-        indexing_device: Optional[Union[enums.Devices, str]] = None,
-        search_device: Optional[Union[enums.Devices, str]] = None,
         api_key: str = None
     ) -> None:
         """
         Parameters
         ----------
         url:
-            The url to the S2Search API (ex: http://localhost:9200)
+            The url to the Marqo instance (ex: http://???)
         """
         self.cluster_is_remote = False
         self.cluster_is_s2search = False
         self.url = self.set_url(url)
         self.timeout = timeout
         self.api_key = api_key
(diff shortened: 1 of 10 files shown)


tests/v0_tests/test_parallel.py
```diff
@@ -19,7 +19,7 @@ def setUp(self) -> None:
         except MarqoApiError as s:
             pass
         
-        self.para_params = {'server_batch_size':10, 'processes':2}
+        self.para_params = {'server_batch_size':10, 'processes':2, 'device': "cpu"}
```
pandu-k, line 22: is this just to make this test a little more stable (given issues with parallel indexing?)

pandu-k (approved):

tests/v0_tests/test_parallel.py
vicilliar: yes, exactly. i did it in `test_logging` as well, since MP with cuda has some weird issues.

Actually i'm still having a slight issue with `test_logging`. When I run those tests alone, they pass, but when I run them as part of the whole test suite, they make marqo break witha  `Broken Pipe Error`. Pretty sure this is some issue with MP.

src/marqo/config.py
```diff
@@ -13,25 +13,19 @@ def __init__(
...
-        indexing_device: Optional[Union[enums.Devices, str]] = None,
-        search_device: Optional[Union[enums.Devices, str]] = None,
         api_key: str = None
     ) -> None:
         """
         Parameters
         ----------
         url:
-            The url to the S2Search API (ex: http://localhost:9200)
+            The url to the Marqo instance (ex: http://???)
```
farshidz: `http://localhost:9200` seems like a better example than `http://???`

farshidz (commented):
In addition to removing default device, which is the purpose of the PR, this is also removing the ability to set devices at client level. It appears sensible to be able to set this at client level and override at method call only if needed.

tests/v0_tests/test_parallel.py
vicilliar: fixed. just reran it again. this may have been caused by running marqo-os locally alongside cuda DIND, meaning i had 2 instances of marqo-os running.

PR #106, opened 2023-06-27
set auto_refresh=false by default

wanliAlex:

we set the default value of auto_refresh in add_documents, update_documents, and delete_documents to False

the default value is true

set to false

yes, this is a breaking change. User need to specify `auto_refresh=True` now.

* **Other information**:

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -239,7 +239,7 @@ class Index:
     def add_documents(
         self,
         documents: List[Dict[str, Any]],
-        auto_refresh=True,
+        auto_refresh=False,
         server_batch_size: int = None,
         client_batch_size: int = None,
         processes: int = None,
@@ -255,9 +255,9 @@ class Index:
 
         Args:
             documents: List of documents. Each document should be a dictionary.
-            auto_refresh: Automatically refresh the index. If you are making
-                lots of requests, it is advised to set this to False to
-                increase performance.
+            auto_refresh: Whether automatically refresh the index or not, default is False.
+                If you need your documents to be available immediately after indexing,
+                set this to True.
             server_batch_size: if it is set, documents will be indexed into batches
                 on the server as they are indexed. Otherwise documents are unbatched
                 server-side.
@@ -291,7 +291,7 @@ class Index:
     def update_documents(
         self,
         documents: List[Dict[str, Any]],
-        auto_refresh=True,
+        auto_refresh = False,
         server_batch_size: int = None,
         client_batch_size: int = None,
         processes: int = None,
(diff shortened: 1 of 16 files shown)


src/marqo/index.py
```diff
@@ -255,9 +255,9 @@ def add_documents(
 
         Args:
             documents: List of documents. Each document should be a dictionary.
-            auto_refresh: Automatically refresh the index. If you are making
-                lots of requests, it is advised to set this to False to
-                increase performance.
+            auto_refresh: Whether automatically refresh the index or not, default is False.
+                If you need your documents to be available immediately after indexing,
+                set this to True.
```
pandu-k, line 260: wording:

Whether automatically refresh the index or not, default is False.
If you need your documents to be available immediately after indexing, set this to True.

-> 
Whether _to_ automatically refresh the index or not. This is False by default. 
If you need your documents to be available immediately after indexing, set this to True. Be careful about setting this  to True if you are doing a large indexing operation, as it can have a significant performance impact. It is better to index with auto_refresh=False, and then calling refresh() after indexing has completed.

src/marqo/index.py
```diff
@@ -452,7 +452,7 @@ def _generic_add_update_docs(
         mq_logger.debug(f"add_documents completed. total time taken: {(total_add_docs_time):.3f}s.")
         return res
 
-    def delete_documents(self, ids: List[str], auto_refresh: bool = None) -> Dict[str, int]:
+    def delete_documents(self, ids: List[str], auto_refresh: bool = False) -> Dict[str, int]:
```
pandu-k, line 455: for delete documents, we want to let Marqo determine the default, rather than having it declared in two places. So let's keep this None by default

pandu-k: We don't have the same luxury (of allowing Marqo to decide) for add_documents, because the Python client needs to know when to make the refresh request (for client batching)

tests/v0_tests/test_add_documents.py
```diff
@@ -618,4 +620,111 @@ def test_multimodal_combination_doc(self):
             self.warm_request(self.client.index(self.index_name_1).search,"")
 
         space_tensor_res = self.client.index(self.index_name_1).search("")
-        assert space_tensor_res["hits"][0]["_id"] == "111"
\ No newline at end of file
+        assert space_tensor_res["hits"][0]["_id"] == "111"
+
```
pandu-k, line 624: These little unit tests are awesome! It really helps us be certain about the behaviour of py-marqo going forward!

pandu-k:
confirmed that tox tests pass

pandu-k:
Note: we need to confirm the positive difference this makes, in Marqo, before merging

PR #107, opened 2023-06-27
Removing support for update docs

pandu-k:

Removing deprecated update method

The update docs endpoint is still available

Removing support for the update docs endpoint

Yes

* **Other information**:
tox unit tests pass

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -280,7 +280,7 @@ class Index:
             non_tensor_fields = []
         if image_download_headers is None:
             image_download_headers = dict()
-        return self._generic_add_update_docs(
+        return self._add_docs_organiser(
             update_method="replace",
             documents=documents, auto_refresh=auto_refresh, server_batch_size=server_batch_size,
             client_batch_size=client_batch_size, processes=processes, device=device, non_tensor_fields=non_tensor_fields,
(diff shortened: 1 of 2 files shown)


pandu-k:
image

src/marqo/index.py
```diff
@@ -280,61 +280,15 @@ def add_documents(
             non_tensor_fields = []
         if image_download_headers is None:
             image_download_headers = dict()
-        return self._generic_add_update_docs(
+        return self._add_docs_organiser(
             update_method="replace",
```
vicilliar: Do we also want to remove the `update_method` parameter? If so, we should remove it from other functions like `_batch_request` as well.

src/marqo/index.py
```diff
@@ -280,61 +280,15 @@ def add_documents(
...
-        if non_tensor_fields is None:
-            non_tensor_fields = []
-        return self._generic_add_update_docs(
-            update_method="update",
-            documents=documents, auto_refresh=auto_refresh, server_batch_size=server_batch_size,
-            client_batch_size=client_batch_size, processes=processes, device=device, non_tensor_fields=non_tensor_fields,
-        )
-
-    def _generic_add_update_docs(
+    def _add_docs_organiser(
```
vicilliar, line 290: Possible suggestion: what are your thoughts on calling this `add_documents_orchestrator` since it has a similar purpose to the orchestrator in marqo?

vicilliar (changes requested):

src/marqo/index.py
pandu-k: I'd prefer to keep the naming conventions decoupled between the two repos

src/marqo/index.py
pandu-k: Good catch - removed update_method params

pandu-k:
confirmed that tox unit tests pass, again

vicilliar (approved):

pandu-k:
tox unit tests pass after merging mainline in

PR #114, opened 2023-07-06
Align `get_indexes()` with API spec

OwenElliottDev:

Bug fix

Currently get_indexes returns a dictionary of str and list of Index objects. The docstring and the REST API however return a dictionary of str and a list of dicts of str and str.

e.g. the python client returns
```
{'results': [<marqo.index.Index object at 0x103f65090>]}
```
Where the REST API returns
```
{'results': [{"index-name", "my-first-index"}]}
```

The new behaviour is to have `get_indexes()` return the same structure as the REST API
```
>>> mq.get_indexes()
{'results': [{"index-name", "my-first-index"}]}
```

This PR does break the use case where users use the `Index` objects returned from `get_indexes()`.

e.g.
```
>>> indexes = mq.get_indexes()['results']
>>> for index in indexes:
...    print(index.get_stats())
>>>
```

However this usage pattern is never documented anywhere or used in any examples as far as I can see. It only appears in one of the tests for the `get_indexes()` method.

* **Other information**:
This changes means that you can print the results as dictionaries have `__repr__`.

An alternative would be to add `__getitem__` and `__repr__` to the index object to retain both the existing functionality and let users treat it like a dict. e.g.
```
>>> indexes = mq.get_indexes()['results']
>>> for index in indexes:
...    print(index.get_stats())
>>> print(indexes)
[Index(config=<marqo.config.Config object at 0x10382f610>, http=<marqo._httprequests.HttpRequests object at 0x102682b10>, index_name='my-first-index', created_at=None, updated_at=None)]
>>> indexes[0]['index_name']
'my-first-index'
```

### DIFF
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -121,18 +121,15 @@ class Client:
             return Index(self.config, index_name=index_name)
         raise Exception('The index UID should not be None')
 
-    def get_indexes(self) -> Dict[str, List[Index]]:
+    def get_indexes(self) -> Dict[str, List[Dict[str, str]]]:
         """Get all indexes.
 
         Returns:
-        Indexes, a dictionary with the name of indexes.
+            Indexes, a dictionary with the name of indexes.
         """
         response = self.http.get(path='indexes')
         response['results'] = [
-            Index(
-                config=self.config,
-                index_name=index_info["index_name"],
-            )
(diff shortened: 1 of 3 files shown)


OwenElliottDev:
@pandu-k what are your thoughts on this? I think we discussed it about 2 months ago but I never got around to taking in further.

PR #115, opened 2023-07-06
MC-35 marqo GA changes

danyilq:

Support for marqo GA.

when url is api.marqo.ai it uses resolver for operations related to indexes to get their url.

No.

* **Other information**:

### DIFF
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -47,10 +47,11 @@ class Client:
         sentence_overlap=0,
         image_preprocessing_method=None,
         settings_dict=None,
-        inference_type="marqo.small",
-        storage_type="marqo.medium",
+        inference_node_type=None,
+        storage_node_type=None,
+        inference_node_count=1,
     ) -> Dict[str, Any]:
-        """Create the index.
+        """Create the index. Please refer to the marqo cloud to see options for inference and storage node types.
 
         Args:
             index_name: name of the index.
@@ -63,8 +64,9 @@ class Client:
             settings_dict: if specified, overwrites all other setting
                 parameters, and is passed directly as the index's
                 index_settings
-            inference_type:
-            storage_type:
+            inference_node_type:
+            storage_node_type:
+            inference_node_count;
         Returns:
             Response body, containing information about index creation result
         """
@@ -74,7 +76,8 @@ class Client:
             model=model, normalize_embeddings=normalize_embeddings,
             sentences_per_chunk=sentences_per_chunk, sentence_overlap=sentence_overlap,
             image_preprocessing_method=image_preprocessing_method,
-            settings_dict=settings_dict, inference_type=inference_type, storage_type=storage_type,
+            settings_dict=settings_dict, inference_node_type=inference_node_type, storage_node_type=storage_node_type,
+            inference_node_count=inference_node_count
         )
 
     def delete_index(self, index_name: str) -> Dict[str, Any]:
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -46,7 +46,9 @@ class Client:
         sentences_per_chunk=2,
         sentence_overlap=0,
         image_preprocessing_method=None,
-        settings_dict=None
+        settings_dict=None,
+        inference_type="marqo.small",
+        storage_type="marqo.medium",
     ) -> Dict[str, Any]:
         """Create the index.
 
@@ -61,6 +63,8 @@ class Client:
             settings_dict: if specified, overwrites all other setting
                 parameters, and is passed directly as the index's
                 index_settings
+            inference_type:
+            storage_type:
         Returns:
             Response body, containing information about index creation result
         """
@@ -70,7 +74,7 @@ class Client:
             model=model, normalize_embeddings=normalize_embeddings,
             sentences_per_chunk=sentences_per_chunk, sentence_overlap=sentence_overlap,
             image_preprocessing_method=image_preprocessing_method,
-            settings_dict=settings_dict
+            settings_dict=settings_dict, inference_type=inference_type, storage_type=storage_type,
         )
 
     def delete_index(self, index_name: str) -> Dict[str, Any]:
@@ -101,7 +105,7 @@ class Client:
         """
         ix = Index(self.config, index_name)
         # verify it exists:
-        self.http.get(path=f"indexes/{index_name}/stats")
+        self.http.get(path=f"indexes/{index_name}/stats", index_name=index_name)
         return ix
 
     def index(self, index_name: str) -> Index:
@@ -170,18 +174,14 @@ class Client:
     def health(self):
         return self.http.get(path="health")
 
-
     def eject_model(self, model_name:str, model_device:str):
         return self.http.delete(path=f"models?model_name={model_name}&model_device={model_device}")
-
 
     def get_loaded_models(self):
         return self.http.get(path="models")
 
-
     def get_cuda_info(self):
         return self.http.get(path="device/cuda")
-
 
     def get_cpu_info(self):
         return self.http.get(path="device/cpu")
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -61,4 +61,4 @@ class Config:
         if self.cluster_is_marqo and not index_name:
             return self.url + "/api"
         self.marqo_url_resolver.refresh_urls_if_needed(index_name)
-        return self.marqo_url_resolver.urls_mapping[index_name]
+        return self.marqo_url_resolver[index_name]
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -53,6 +53,9 @@ class Config:
         return self.url
 
     def get_url(self, index_name=None,):
+        """Get the URL, and infers whether that url is marqo cloud,
+        and if it is targeting a specific index,
+        kicks off the url resolver refresh process if needed"""
         if not self.cluster_is_marqo:
             return self.url
         if self.cluster_is_marqo and not index_name:
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -2,6 +2,7 @@ from typing import Optional, Union
 from marqo import enums, utils
 import urllib3
 import warnings
+from marqo.marqo_url_resolver import MarqoUrlResolver
 
 
 class Config:
@@ -24,9 +25,11 @@ class Config:
         """
         self.cluster_is_remote = False
         self.cluster_is_s2search = False
+        self.cluster_is_marqo = False
+        self.marqo_url_resolver = None
+        self.api_key = api_key
         self.url = self.set_url(url)
         self.timeout = timeout
-        self.api_key = api_key
         # suppress warnings until we figure out the dependency issues:
         # warnings.filterwarnings("ignore")
         self.use_telemetry = use_telemetry
@@ -43,5 +46,15 @@ class Config:
             self.cluster_is_remote = True
             if "s2search.io" in lowered_url:
                 self.cluster_is_s2search = True
+            if "api.marqo.ai" in lowered_url:
+                self.cluster_is_marqo = True
+                self.marqo_url_resolver = MarqoUrlResolver(api_key=self.api_key, expiration_time=60)
         self.url = url
         return self.url
+
+    def get_url(self, index_name=None,):
+        if not self.cluster_is_marqo:
+            return self.url
+        if self.cluster_is_marqo and not index_name:
+            return self.url + "/api"
+        return self.marqo_url_resolver.urls_mapping[index_name]
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -54,11 +54,10 @@ class Config:
 
     def get_url(self, index_name=None,):
         """Get the URL, and infers whether that url is marqo cloud,
-        and if it is targeting a specific index,
-        kicks off the url resolver refresh process if needed"""
+        and if it is targeting a specific index resolves the index-specific url"""
         if not self.cluster_is_marqo:
             return self.url
         if self.cluster_is_marqo and not index_name:
             return self.url + "/api"
-        self.marqo_url_resolver.refresh_urls_if_needed(index_name)
+        # calls resolver to get index-specific url for when cluster is marqo and index_name is not None
         return self.marqo_url_resolver[index_name]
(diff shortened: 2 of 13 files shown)


pandu-k:
hi @danyilq ! Have you ran `tox` unit tests? And do they pass?

src/marqo/client.py
```diff
@@ -46,7 +46,9 @@ def create_index(
         sentences_per_chunk=2,
         sentence_overlap=0,
         image_preprocessing_method=None,
-        settings_dict=None
+        settings_dict=None,
+        inference_type="marqo.small",
+        storage_type="marqo.medium",
```
pandu-k: Are these defaults configured in the cloud somewhere? If so, I think we should leave these as `None` by default, and let the cloud determine defaults if not explicitly set by the user. 

Otherwise we will have defaults defined in 2 places.

pandu-k: Also, how about being even more transparent and explicit about what these parameters mean? How about
`inference_node_type` and `storage_node_type`

src/marqo/config.py
```diff
@@ -43,5 +46,15 @@ def set_url(self, url):
             self.cluster_is_remote = True
             if "s2search.io" in lowered_url:
                 self.cluster_is_s2search = True
+            if "api.marqo.ai" in lowered_url:
+                self.cluster_is_marqo = True
+                self.marqo_url_resolver = MarqoUrlResolver(api_key=self.api_key, expiration_time=60)
         self.url = url
         return self.url
+
+    def get_url(self, index_name=None,):
```
pandu-k, line 55: Add a docstring about what this is for

src/marqo/index.py
```diff
@@ -54,7 +56,9 @@ def create(config: Config, index_name: str,
                sentences_per_chunk=2,
                sentence_overlap=0,
                image_preprocessing_method=None,
-               settings_dict: dict = None
+               settings_dict: dict = None,
+               inference_type: str = "marqo.CPU",
```
pandu-k: why is both `"inference_type": "marqo.small",` and `inference_type: str = "marqo.CPU",` defined in the code in different places?

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,29 @@
+import time
+
+import requests
+
+
+class MarqoUrlResolver:
```
pandu-k, line 8: Needs tests for each method, and tests around the changing object state.

src/marqo/index.py
```diff
@@ -91,7 +97,17 @@ def create(config: Config, index_name: str,
...
+                return req.post(f"indexes/{index_name}", body=cl_settings)
+            cl_settings['inference_type'] = inference_type
+            cl_settings['storage_type'] = storage_type
+            req.post(f"indexes/{index_name}", body=cl_settings)
+            creation = req.get(f"indexes/{index_name}/status")
+            while creation['index_status'] != 'READY':
+                time.sleep(10)
+                creation = req.get(f"indexes/{index_name}/status")
+                mq_logger.info(f"Index creation status: {creation['status']}")
+            return creation
```
pandu-k: Needs tests for creating indexes for the cloud

src/marqo/config.py
pandu-k: Simple tests about the behaviour would be good to prevent accidental future regressions

src/marqo/config.py
```diff
@@ -43,5 +46,15 @@ def set_url(self, url):
...
+        if not self.cluster_is_marqo:
+            return self.url
+        if self.cluster_is_marqo and not index_name:
+            return self.url + "/api"
+        return self.marqo_url_resolver.urls_mapping[index_name]
```
pandu-k: do we need error handling if `index_name` not in `urls_mapping`

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, api_key=None, expiration_time: int = 60):
+        """ URL Resolver is a cache for urls that are resolved to their respective indices only for marqo cloud. """
+        self.timestamp = time.time() - 60
```
pandu-k: consider `self.timestamp = time.time() - expiration_time ` if we consistently want to refresh the URLs after initialisation

src/marqo/config.py
pandu-k: We should also note in the docstring that this method kicks off the URL refresh process

pandu-k (commented):
One of the problems with this is that `HttpRequests. send_request()` (which is called on every HTTP request to Marqo - including latency critical search requests), periodically starts blocking calls to `refresh_urls_if_needed` before the  requested HTTP request be sent.

One way around this to make `refresh_urls_if_needed` execute in its own thread, unless the requested index is not found in `_urls_mapping`, in which case a blocking call is alright

src/marqo/index.py
```diff
@@ -74,8 +75,9 @@ def create(config: Config, index_name: str,
             settings_dict: if specified, overwrites all other setting
                 parameters, and is passed directly as the index's
                 index_settings
-            inference_type: inference type for the index
-            storage_type: storage type for the index
+            inference_node_type: inference type for the index
+            storage_node_type: storage type for the index
+            inference_node_count: number of inference nodes for the index
```
pandu-k, line 80: Can you please create unit tests for `inference_node_type`, `inference_node_count` and `storage_node_type` parameters, including edge cases (`None` or `0`)

tests/v0_tests/test_index.py
```diff
@@ -200,4 +200,13 @@ def test_create_custom_number_of_replicas(self):
         self.client.create_index(index_name=self.index_name_1, settings_dict = settings)
         index_setting = self.client.index(self.index_name_1).get_settings()
         print(index_setting)
-        assert intended_replicas == index_setting['number_of_replicas']
\ No newline at end of file
+        assert intended_replicas == index_setting['number_of_replicas']
+
+    @mock.patch("marqo._httprequests.HttpRequests.post", return_value="Record added to DynamoDB table successfully")
+    @mock.patch("marqo._httprequests.HttpRequests.get", return_value={"index_status": "READY"})
+    def test_create_marqo_cloud_index(self, mock_post, mock_get):
```
pandu-k: check that URL string params are as expected

src/marqo/index.py
```diff
@@ -54,7 +56,10 @@ def create(config: Config, index_name: str,
                sentences_per_chunk=2,
                sentence_overlap=0,
                image_preprocessing_method=None,
-               settings_dict: dict = None
+               settings_dict: dict = None,
+               inference_node_type: str = None,
+               storage_node_type: str = None,
+               inference_node_count: int = 1,
```
pandu-k, line 62: should also let the API decide what the default `inference_node_count` is?

danyilq: this value must be greater than 0 anyways, api will validate the value, but I think that having it set to 1 in the client is reasonable so that customer doesn't get unexpected value by api. Whilst forcing customer to choose specific storage and inference types.

src/marqo/config.py
pandu-k: There's no guarantee that the index_name is going to end up in mappings right? So for a non existent index, they won't get a `404 index not found` from Marqo,  they will get a KeyError

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,30 @@
...
+class MarqoUrlResolver:
+    def __init__(self, api_key=None, expiration_time: int = 60):
+        """ URL Resolver is a cache for urls that are resolved to their respective indices only for marqo cloud. """
+        self.timestamp = time.time() - expiration_time
+        self._urls_mapping = {}
+        self.api_key = api_key
+        self.expiration_time = expiration_time
+
+    def refresh_urls_if_needed(self, index_name):
+        if index_name not in self._urls_mapping and time.time() - self.timestamp > self.expiration_time:
```
pandu-k: This basically means that we are OK with an index_name's url _never_ being updated in this client once set (including if it has been deleted and recreated). Are we sure we are alright with this?

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,30 @@
+import time
+
+import requests
+
+
+class MarqoUrlResolver:
+    def __init__(self, api_key=None, expiration_time: int = 60):
```
pandu-k: nit: Are we sure that we are willing to only check for updates after a minute? This feels a little long. For example, someone else creates an index, I will have to wait up to a minute after it is available before I can use it. Perhaps a lower value, like 15 seconds or so?

danyilq: It sounds reasonable, I agree

src/marqo/marqo_url_resolver.py
danyilq: how about we refresh it once in 5-15 minutes if index is already presented, if customer were to delete and recreate their deletion would take around 5 minutes and after <=(5-15) minutes their index will be deleted from urls list and then each time they request it would refresh at the same rate as we have default.

src/marqo/config.py
danyilq: Mindset was to keep client dumb and let api handle it, but in that case we indeed need to handle it with client, wip on that one

danyilq: Is it reasonable to raise Exception(f"Index {item} is still being created") when index is creating
and Exception(f"Index {item} does not exist") or should it be softer and not stopping the code?

pandu-k: Yes I agree that we need to keep the client as simple as possible, but we may have no choice here. 

In terms of error message we should be as clear as possible.

For example something like this (feel free to modify this if needed) :

"The Python client could not resolve the endpoint for the index name `my-index`. This could be due to the index is still in the process of being created, or that the client's cache has not yet been updated. 
- please try again in a couple of minutes, to allow the client's cache to be updated
- you can also try the querying the index's status: GET https://<INSERT INDEX ENDPOINT HERE>/status
- If the issue still persists please make a Marqo cloud support ticket here <LINK

pandu-k: Yes I agree that we need to keep the client as simple as possible, but we may have no choice here. 

In terms of error message we should be as clear as possible.

For example something like this (feel free to modify this if needed) :
``` 
The Python client could not resolve the endpoint for the index name `my-index`. 
This could be due to the index is still in the process of being created, or that the client's cache has not yet been updated. 
- please try again in a couple of minutes, to allow the client's cache to be updated
- you can also try the querying the index's status: `curl https://<INSERT INDEX ENDPOINT HERE>/status`
- If the issue still persists please make a Marqo cloud support ticket here <LINK>
```

src/marqo/errors.py
```diff
@@ -165,3 +165,27 @@ class BackendTimeoutError(InternalError):
...
+    """Error when Marqo index is not ready"""
+    code = "index_not_ready_cloud"
+    status_code = HTTPStatus.NOT_FOUND
+
+    def __init__(self, index_name: str,) -> None:
+        self.message = f"The Python client could not resolve the endpoint for the index name {index_name}. " \
+                       f"This could be due to the index is still in the process of being created," \
+                       f" or that the client's cache has not yet been updated.\n" \
+                       f"- Please try again in a couple of minutes, " \
+                       f"or you can query the index status with get_status function to see when index is ready.\n" \
```
pandu-k: `mq.('my-index').get_status()`

pandu-k: - Consider subclassing this, because it's a client specific, rather than HTTP errror: https://github.com/marqo-ai/py-marqo/blob/ab2c3de94f02ad0ef8440da743f6e18600f701a1/src/marqo/errors.py#L6

src/marqo/errors.py
```diff
@@ -165,3 +165,27 @@ class BackendTimeoutError(InternalError):
     def __init__(self, message: str,) -> None:
         self.message = f"Timeout error communicating with Marqo: {message}"
 
+
+class MarqoIndexNotReadyError(__InvalidRequestError):
```
pandu-k: Add `Cloud` to error name: `MarqoIndexNotReadyErrorCloud`

src/marqo/errors.py
```diff
@@ -165,3 +165,27 @@ class BackendTimeoutError(InternalError):
...
+    def __init__(self, index_name: str,) -> None:
+        self.message = f"The Python client could not resolve the endpoint for the index name {index_name}. " \
+                       f"This could be due to the index is still in the process of being created," \
+                       f" or that the client's cache has not yet been updated.\n" \
+                       f"- Please try again in a couple of minutes, " \
+                       f"or you can query the index status with get_status function to see when index is ready.\n" \
+                       f"- If the problem persists, please contact marqo support at support@marqo.ai"
+
+
+class MarqoIndexNotFoundError(__InvalidRequestError):
```
pandu-k: ditto RE "Add Cloud to error name"

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,38 @@
...
+        self.refresh_urls_if_needed(item)
+        if item in self._urls_mapping['READY']:
+            return self._urls_mapping['READY'][item]
+        if item in self._urls_mapping['CREATING']:
+            raise MarqoIndexNotReadyError(item)
+        raise MarqoIndexNotFoundError(item)
+
+    def _refresh_urls(self):
+        response = requests.get('https://api.marqo.ai/api/indexes',
+                                headers={"x-api-key": self.api_key}).json()
```
pandu-k: Consider logging a warning if response is not OK

src/marqo/marqo_url_resolver.py
```diff
@@ -0,0 +1,38 @@
...
+        self.api_key = api_key
+        self.expiration_time = expiration_time
+
+    def refresh_urls_if_needed(self, index_name):
+        if index_name not in self._urls_mapping['READY'] and time.time() - self.timestamp > self.expiration_time:
+            # fast refresh to catch if index was created
+            self._refresh_urls()
+        if index_name in self._urls_mapping['READY'] and time.time() - self.timestamp > 360:
+            # slow refresh in case index was deleted
+            self._refresh_urls()
```
pandu-k: Consider adding a timeout here, for the slow case (we don't want to hold up search/add docs requests if the control plane is acting badly). Perhaps 3 seconds

pandu-k (commented):
Nice!
+ make a few tests around index_name within HTTP requests (to prevent future regressions)
  + this is because it isn't intuitive that `http.get(url, index_name=index_name)` triggers Marqo-cloud specific functionality, when `index_name` is set
  + You may want to add a docstring comment to explain this non-intuitive functionality 
+ tests that it doesn't use the API resolver if the URL isn't api.marqo

pandu-k (approved):

PR #123, opened 2023-07-16
Add tensor_fields parameter to add_documents

farshidz:

Feature

To add new documents to an index, `add_documents` receives `non_tensor_fields` and treats all fields except those in `non_tensor_fields` as tensor fields.

`add_documents` will now accept `tensor_fields`. When this is provided, all fields not in `tensor_fields` are treated at non-tensor fields. This is in line with Marqo API changes in this PR https://github.com/marqo-ai/marqo/pull/538

`non_tensor_fields` is now marked as deprecated and using it will cause a warning, indicating that the field will be removed in Marqo 2.0.0.

Yes. Not specifying `non_tensor_fields` previously led to all fields being treated as tensor fields. This will now lead to an error. 

* **Other information**:

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -268,6 +268,7 @@ class Index:
         auto_refresh: bool = True,
         client_batch_size: int = None,
         device: str = None,
+        tensor_fields: List[str] = None,
         non_tensor_fields: List[str] = None,
         use_existing_tensors: bool = False,
         image_download_headers: dict = None,
@@ -287,7 +288,11 @@ class Index:
                 client-side.
             device: the device used to index the data. Examples include "cpu",
                 "cuda" and "cuda:2"
-            non_tensor_fields: fields within documents to not create and store tensors against.
+            tensor_fields: fields within documents to create and store tensors against.
+            non_tensor_fields: fields within documents to not create and store tensors against. Cannot be used with
+                tensor_fields.
+                .. deprecated:: 2.0.0
+                    This parameter has been deprecated and will be removed in Marqo 2.0.0. User tensor_fields instead.
             use_existing_tensors: use vectors that already exist in the docs.
             image_download_headers: a dictionary of headers to be passed while downloading images,
                 for URLs found in documents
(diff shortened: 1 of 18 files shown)


src/marqo/index.py
```diff
@@ -296,15 +301,26 @@ def add_documents(
...
+                                         '`non_tensor_fields` has been deprecated and will be removed in Marqo 2.0.0. '
+                                         'Its use is discouraged.')
+
+        if tensor_fields is None and non_tensor_fields is None:
+            raise errors.InvalidArgError('You must include the `tensor_fields` parameter. '
+                                         'Use `tensor_fields=[]` to index for lexical-only search.')
+
+        if non_tensor_fields is not None:
+            logging.warning('The `non_tensor_fields` parameter has been deprecated and will be removed in Marqo 2.0.0. '
+                            'Use `tensor_fields` instead.')
```
pandu-k: I recommend using the client's logger: `mq_logger.warning(...` so that users can easily control the log level of the entire client

src/marqo/index.py
```diff
@@ -336,12 +355,15 @@ def _add_docs_organiser(
...
-            "nonTensorFields" : non_tensor_fields,
             "useExistingTensors" : use_existing_tensors,
             "imageDownloadHeaders" : image_download_headers,
             "mappings" : mappings,
             "modelAuth": model_auth,
         }
+        if tensor_fields is not None:
+            base_body['tensorFields'] = tensor_fields
+        else:
+            base_body['nonTensorFields'] = non_tensor_fields
```
pandu-k, line 366: I thought `nonTensorFields` has now become a url query string param?

pandu-k:
+ consider an creating an assertlogs test that confirms that we throw warnings for non_tensor_fields

src/marqo/index.py
farshidz: `non_tensor_fields` is a deprecated query parameter that the client doesn't use but the API still supports. Current behaviour is to have these in the body. So `tensorFields` is only supported in the body.

pandu-k (approved):

PR #129, opened 2023-07-25
Add index heath check API

wanliAlex:

add client.index(index_name).health() API

we use client.health() API

raising deprecation warning when using client.health() API and add client.index(index_name).health() API.

We introduce a new API. But it's still backward-compatible.

* **Other information**:

### DIFF
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -52,18 +52,18 @@ class Client:
         self._marqo_minimum_supported_version_check()
 
     def create_index(
-        self, index_name: str,
-        treat_urls_and_pointers_as_images=False, model=None,
-        normalize_embeddings=True,
-        sentences_per_chunk=2,
-        sentence_overlap=0,
-        image_preprocessing_method=None,
-        settings_dict=None,
-        inference_node_type=None,
-        storage_node_type=None,
-        inference_node_count=1,
-        storage_node_count=1,
-        replicas_count=0,
+            self, index_name: str,
+            treat_urls_and_pointers_as_images=False, model=None,
+            normalize_embeddings=True,
+            sentences_per_chunk=2,
+            sentence_overlap=0,
+            image_preprocessing_method=None,
+            settings_dict=None,
+            inference_node_type=None,
+            storage_node_type=None,
+            inference_node_count=1,
+            storage_node_count=1,
+            replicas_count=0,
     ) -> Dict[str, Any]:
         """Create the index. Please refer to the marqo cloud to see options for inference and storage node types.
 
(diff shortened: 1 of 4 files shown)


Mehul-Porwal (approved):

src/marqo/client.py
```diff
@@ -192,6 +192,9 @@ def get_marqo(self):
         return self.http.get(path="")
 
     def health(self):
+        mq_logger.warning('The `client.health()` API has been deprecated and will be removed in '
+                          'Marqo 2.0.0. Use `client.index(index_name).health()` instead. '
+                          'Check `https://docs.marqo.ai/latest/API-Reference/indexes/` for more details.')
```
pandu-k, line 197: What does this endpoint look like if you are using the cloud? Does it result in a JSON decode error? 

If so, perhaps it would be better to add some logic that checks if the URL corresponds to the cloud, and then give a more useful error (perhaps after catching the JSON decode error)

src/marqo/client.py
```diff
@@ -195,7 +195,14 @@ def health(self):
         mq_logger.warning('The `client.health()` API has been deprecated and will be removed in '
                           'Marqo 2.0.0. Use `client.index(index_name).health()` instead. '
                           'Check `https://docs.marqo.ai/latest/API-Reference/indexes/` for more details.')
-        return self.http.get(path="health")
+        try:
+            return self.http.get(path="health")
+        except (MarqoWebError, RequestException, TypeError, KeyError) as e:
+            raise errors.BadRequestError("Marqo encountered an error trying to check the health of the Marqo instance. "
+                                         "It is likely you are trying to check the health on Marqo Cloud V2. "
```
pandu-k: Let's make less assumptions in our language: 
`"It is likely you are trying to check the health` -> `If you are trying to check the health`

pandu-k: `Marqo Cloud V2` is internal, please just use `Marqo Cloud`

src/marqo/client.py
```diff
@@ -195,7 +195,14 @@ def health(self):
...
+                                         "The `client.health()` API is not supported on Marqo Cloud V2 and will be removed in"
+                                         "Marqo 2.0.0. Please Use `client.index(index_name).health()` instead. "
```
pandu-k: Can we be careful about spaces. This will become `inMarqo 2.0.0`

Also, for the example, let's make it clearer (as index_name is a local variable, users might get confused):
`client.index("my-index").health()`

src/marqo/client.py
```diff
@@ -195,7 +195,14 @@ def health(self):
...
+                                         "Check `https://docs.marqo.ai/latest/API-Reference/indexes/` for more details.")
```
pandu-k: I prefer using a specific version (rather than latest), so the URL should always work:

`https://docs.marqo.ai/1.0.0/API-Reference/indexes/`

pandu-k (approved):

PR #132, opened 2023-07-31
introduce instance mappings

danyilq:

Introduces new class MarqoInstanceMappings that allows user to specify url for each index to use for index-specific api requests. Moves certain methods from Client class to Index class and deprecates those used in Client

* **Other information**:


src/marqo/client.py
```diff
@@ -161,6 +166,10 @@ def get_indexes(self) -> Dict[str, List[Index]]:
         ]
         return response
 
+    @deprecated(
+        "This method is deprecated and will be removed in Marqo 2.0.0"
+        ", instead use 'client.index(index_name).enrich()"
+    )
     def enrich(self, documents: List[Dict], enrichment: Dict, device: str = None, ):
```
pandu-k: just delete this - no one uses it

src/marqo/_httprequests.py
```diff
@@ -35,9 +35,11 @@ def _operation(self, method: HTTP_OPERATIONS) -> Callable:
 
     def _construct_path(self, path: str, index_name="") -> str:
         """Augment the URL request path based if telemetry is required."""
-        url = f"{self.config.get_url(index_name=index_name)}/{path}"
+        base_url = self.config.instance_mapping.get_url(index_name=index_name) if index_name \
+            else self.config.instance_mapping.get_control_url()
```
pandu-k: Shouldn't the instance_mapping determine what base_url is, rather than having this logic here? 

Also, telemetry is only used for search, bulk_search and add_documents requests, so we shouldn't need the control endpoint here

src/marqo/client.py
```diff
@@ -212,9 +212,13 @@ def _base64url_encode(
         ", instead use 'client.index(index_name).get_marqo()"
     )
     def get_marqo(self):
+        if isinstance(self.config.instance_mapping, MarqoCloudInstanceMappings):
```
farshidz: You can use `Client.is_marqo_cloud` for this. Applies to all others below

src/marqo/client.py
```diff
@@ -232,59 +236,76 @@ def health(self):
...
     @deprecated(
         "This method is deprecated and will be removed in Marqo 2.0.0"
         ", instead use 'client.index(index_name).get_loaded_models()"
     )
     def get_loaded_models(self):
+        if isinstance(self.config.instance_mapping, MarqoCloudInstanceMappings):
+            self.raise_error_for_cloud()
         return self.http.get(path="models")
 
     @deprecated(
```
farshidz, line 248: This annotation seems to support `reason` and `version`. We should use those. https://pypi.org/project/Deprecated/

src/marqo/client.py
```diff
@@ -232,59 +236,76 @@ def health(self):
...
         ", instead use 'client.index(index_name).get_cpu_info()"
     )
     def get_cpu_info(self):
+        if isinstance(self.config.instance_mapping, MarqoCloudInstanceMappings):
+            self.raise_error_for_cloud()
         return self.http.get(path="device/cpu")
 
     def _marqo_minimum_supported_version_check(self):
         min_ver = minimum_supported_marqo_version()
+        url = self.config.instance_mapping.get_control_url()
```
farshidz: This won't work because the control plane won't serve this endpoint. Different data planes could be running different versions so this is a tricky one. I'll get back to you on this.

src/marqo/client.py
```diff
@@ -204,48 +218,81 @@ def health(self):
...
         except (MarqoWebError, RequestException, TypeError, KeyError) as e:
             mq_logger.warning(skip_warning_message)
-            marqo_url_and_version_cache[self.url] = "_skipped"
+            marqo_url_and_version_cache[url] = "_skipped"
         return
+
+    @staticmethod
+    def raise_error_for_cloud(function_name: str = None):
+        raise errors.BadRequestError(
+            f"The `client.{function_name}()` API is not supported on Marqo Cloud and will be removed in "
```
farshidz: I would rephrase the error as 
`mq.{function_name}()` is not supported on Marqo Cloud. Please use `mq.index(...).{function_name}()` instead. Check https://docs.marqo.ai/1.1.0/API-Reference/indexes/ for more details.

Note no need to repeat it'll be removed in 2.0.0 as the deprecation warning covers that.

farshidz (changes requested):

src/marqo/instance_mappings.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC, abstractmethod
+
+from typing import Dict, Optional
+
+
+class InstanceMappings(ABC):
```
pandu-k, line 6: We need to add a clear message here that this doesn't support cases where you have the same index name on more than 1 URL cluster

src/marqo/instance_mappings.py
```diff
@@ -0,0 +1,26 @@
...
+    @abstractmethod
+    def get_url(self, index_name: str) -> str:
```
pandu-k: let's make this more descriptive: how about something like `get_index_base_url`

pandu-k: and other one should be `get_control_base_url`

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -1,37 +1,48 @@
...
 )
+from marqo.instance_mappings import InstanceMappings
+from marqo.marqo_logging import mq_logger
+
 
+class MarqoCloudInstanceMappings(InstanceMappings):
+    _MARQO_CLOUD_URL = "https://api.marqo.ai"
 
-class MarqoUrlResolver:
     def __init__(self, api_key=None, expiration_time: int = 15):
```
pandu-k: We will need more context, and perhaps a better name for the `expiration_time` parameter. It is not clear what this parameter does when we instantiate this class.

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -12,20 +12,20 @@ def test_refresh_urls_if_needed(self, mock_get):
...
         # Wait for more than the expiration time
         time.sleep(0.1)
 
-        resolver.refresh_urls_if_needed("index1")
+        mapping._refresh_urls_if_needed("index1")
 
         # Check that the timestamp has been updated
-        print(resolver.timestamp, initial_timestamp)
-        assert resolver.timestamp > initial_timestamp
+        print(mapping.timestamp, initial_timestamp)
```
pandu-k: rm print statements

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -1,37 +1,48 @@
...
-        """ URL Resolver is a cache for urls that are resolved to their respective indices only for marqo cloud. """
         self.timestamp = time.time() - expiration_time
```
pandu-k: Can we also rename `self.timestamp` to something more descriptive? For example, something like `self.latest_index_mappings_refresh_timestamp`

pandu-k (commented):
We need tests to ensure all the new index-level endpoints work:
Index.get_loaded_models(), etc.
Feel free to copy/paste/modify the existing tests to make this faster

src/marqo/marqo_cloud_instance_mappings.py
danyilq: agreed

danyilq:
> We need tests to ensure all the new index-level endpoints work: Index.get_loaded_models(), etc. Feel free to copy/paste/modify the existing tests to make this faster

@pandu-k Those are already implemented as tests and were tested they work correctly both on V2 on local setup

src/marqo/marqo_cloud_instance_mappings.py
danyilq: @pandu-k chatGPT proposed url_cache_duration it sounds good to me

src/marqo/client.py
```diff
@@ -41,15 +44,24 @@ def __init__(
         url:
             The url to the S2Search API (ex: http://localhost:8882)
```
pandu-k: Let's update the docstring. Here are some recommendations: 
1. Let's use this opportunity to remove the reference to s2search (it should be url to Marqo)
2. Add an explanation of the `instance_mappings` param
3. Add an explanation of the `api_key` param
4. For the explanation of the `url` param, please add a comment here about using the MARQO_CLOUD_URL env var

src/marqo/client.py
```diff
@@ -188,10 +191,24 @@ def _base64url_encode(
...
+    )
     def get_marqo(self):
+        if self.config.is_marqo_cloud:
+            self.raise_error_for_cloud("get_marqo")
         return self.http.get(path="")
 
+    @deprecated(
+        "This method is deprecated and will be removed in Marqo 2.0.0. "
+        "Please use `client.index(index_name).health()` instead. "
+        "Check `https://docs.marqo.ai/latest/API-Reference/indexes/` for more details."
```
pandu-k: for future compatibility, let's pin this to a specific version, for example: `https://docs.marqo.ai/1.1.0/API-Reference/indexes/`

src/marqo/client.py
```diff
@@ -41,15 +44,24 @@ def __init__(
...
+        if url is not None and instance_mappings is not None:
+            raise ValueError("Cannot specify both url and instance_mappings")
+
+        is_marqo_cloud = False
+        if url is not None:
+            if url.lower().startswith(os.environ.get("MARQO_CLOUD_URL", "https://api.marqo.ai")):
+                instance_mappings = MarqoCloudInstanceMappings(control_base_url=url, api_key=api_key)
+                is_marqo_cloud = True
+            else:
+                instance_mappings = DefaultInstanceMappings(url, main_user, main_password)
```
pandu-k, line 62: do we need to set `is_marqo_cloud` to False here?

src/marqo/index.py
```diff
@@ -85,10 +93,21 @@ def create(config: Config, index_name: str,
         Returns:
             Response body, containing information about index creation result
         """
+        def cloud_wait_for_index_ready():
```
pandu-k: Let's not use nested functions where possible. Perhaps we create file `cloud_helpers.py` for these types of functions. Then we can also create unit tests for these functions

src/marqo/instance_mappings.py
```diff
@@ -0,0 +1,33 @@
...
+from typing import Dict, Optional
+
+
+class InstanceMappings(ABC):
+    """
+    Abstract class for mapping index names to URLs.
+
+    Implementations of this class must ensure at least asymptotically constant computational complexity for all methods.
+    An inefficient implementation of this class can cause the Marqo client to be slow.
+    Index names must be unique across all instances.
```
pandu-k: Let's try to be more helpful:

"Index names must be unique across all instances." -> "The namespace of index names is with respect to a Client instance. Index names must be unique across all Marqo URLs referenced in the Mappings. If you want to support the same index name on different Marqo URLs, please use separate Client objects. "

src/marqo/instance_mappings.py
```diff
@@ -0,0 +1,33 @@
...
+
+from typing import Dict, Optional
+
+
+class InstanceMappings(ABC):
+    """
+    Abstract class for mapping index names to URLs.
+
+    Implementations of this class must ensure at least asymptotically constant computational complexity for all methods.
+    An inefficient implementation of this class can cause the Marqo client to be slow.
```
pandu-k, line 11: @farshidz we had an idea for a clarifying sentence after this first sentence ("... as the number of requests per second grow, methods of this implementations of this class should tend towards constant time, given typical usage patterns for a target usecase."). 

Good if @farshidz could fill in details/confirm.

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+        client = marqo.Client(**self.client_settings)
+        index_settings = index_defaults if index_defaults else {}
+        index_settings.update({
+            "inference_type": "marqo.CPU", "storage_class": "marqo.basic", "model": "hf/all_datasets_v4_MiniLM-L6"
+        })
+        while True:
+            if client.http.get(f"/indexes/{index_name}/status") == "READY":
+                break
+            try:
+                client.create_index(index_name, settings_dict=index_settings)
```
pandu-k: why do we try/break/except in a `while True` loop to create an index? 

Shouldn't `Client.create_index` method have logic for waiting on an index to be spun up?

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
         This solves the occurence of tests failing due to eventual consistency implemented in marqo cloud.
         '''
         for i in range(5):
-            func(*args, **kwargs) 
+            func(*args, **kwargs)
+
+    def create_cloud_index(self, index_name, index_defaults=None, **kwargs):
```
pandu-k: we have to be careful here. If the user specifies index_defaults, they would expect an index to be made with those settings. Because we just return if that index exists, the user may start to index that they don't suspect.

At the very least we need to get the index's settings, and to a comparison the index defaults, raising an error if they don't match.

This behaviour then needs to be clearly specified in the docstring

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+    def create_index(self, index_name, index_defaults=None, **kwargs):
+        client = marqo.Client(**self.client_settings)
+        if client.config.is_marqo_cloud:
+            self.create_cloud_index(index_name, index_defaults, **kwargs)
+            return
+        client.create_index(index_name, settings_dict=index_defaults, **kwargs)
+
+    def delete_documents(self, index_name):
+        client = marqo.Client(**self.client_settings)
+        indexes = client.get_indexes()
```
pandu-k, line 262: doesn't this need to be `client.get_indexes()["results"]`? If so, how did you test this without detecting this bug?

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+        if index_name in indexes:
+            if client.http.get(f"/indexes/{index_name}/status") == "READY":
+                try:
+                    idx = client.index(index_name)
+                    docs_to_delete = [i['_id'] for i in idx.search("")['hits']]
```
pandu-k: we need to do this in a for loop probably if number of docs in index > limit. We can check if index.get_stats()["numbeOfDocuments"] == 0, and then break

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+                    if docs_to_delete:
+                        idx.delete_documents(docs_to_delete)
```
pandu-k: Do `idx.delete_documents(docs_to_delete, auto_refresh=True) ` to make this delete a blocking call

tests/v0_tests/test_add_documents.py
```diff
@@ -14,35 +14,43 @@
...
 
 class TestAddDocuments(MarqoTestCase):
-
     def setUp(self) -> None:
         self.client = Client(**self.client_settings)
         self.index_name_1 = "my-test-index-1"
-        try:
-            self.client.delete_index(self.index_name_1)
-        except MarqoApiError as s:
-            pass
```
pandu-k, line 24: This is frequently repeated code. Can we move this to the superclass?

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+                    continue
+
+    def create_index(self, index_name, index_defaults=None, **kwargs):
+        client = marqo.Client(**self.client_settings)
+        if client.config.is_marqo_cloud:
+            self.create_cloud_index(index_name, index_defaults, **kwargs)
+            return
+        client.create_index(index_name, settings_dict=index_defaults, **kwargs)
+
+    def delete_documents(self, index_name):
```
pandu-k: Let's remove ambiguity about what this function does. We also want to remove confusion between this method and the Index's method of the same name. Perhaps we name this `delete_all_docs`

tests/marqo_test.py
```diff
@@ -146,4 +142,39 @@ def warm_request(self, func, *args, **kwargs):
...
+        while True:
+            if client.http.get(f"/indexes/{index_name}/status") == "READY":
+                break
+            try:
+                client.create_index(index_name, settings_dict=index_settings)
+            except marqo.errors.MarqoWebError as e:
+                if e.status_code == 409:
+                    continue
+
+    def create_index(self, index_name, index_defaults=None, **kwargs):
```
pandu-k: We want to remove confusion between this method and the Index's method of the same name. Perhaps we name this `create_test_index`

tests/v0_tests/test_index.py
```diff
@@ -40,9 +49,9 @@ def test_create_index_settings_dict(self):
             mock__post = mock.MagicMock()
             @mock.patch("marqo._httprequests.HttpRequests.post", mock__post)
             def run():
-                self.client.create_index(
+                self.create_index(
                     index_name=self.index_name_1,
-                    settings_dict=settings_dict,
+                    index_defaults=settings_dict,
```
pandu-k: the `index_defaults` parameter in `create_index` is misleading because, at least in this case it's actually just the index settings (with the index_defaults being a field within the index_settings dict)

pandu-k: the `index_defaults` parameter in `create_index` needs to be more clearly defined.

tests/v0_tests/test_index.py
```diff
@@ -292,6 +301,95 @@ def test_create_marqo_cloud_index_wrong_inference_node_count(self, mock_get, moc
             },
             'number_of_shards': 1, 'number_of_replicas': 0,
             'inference_type': "marqo.CPU", 'storage_class': "marqo.basic", 'number_of_inferences': -1})
-        mock_get.assert_called_with(path="indexes/my-test-index-1/status")
+        mock_get.assert_called_with("indexes/my-test-index-1/status")
         assert result == {"error": "inference_node_count must be greater than 0"}
```
pandu-k, line 311: Can we do a test where we have multiple clients, and multiple indexes created per client?

pandu-k: We may need to mock stuff to make this test viable on the cloud ^

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -0,0 +1,120 @@
+import time
+from unittest.mock import patch
+
+from marqo.marqo_cloud_instance_mappings import MarqoCloudInstanceMappings
+from tests.marqo_test import MarqoTestCase
+from marqo.errors import MarqoCloudIndexNotFoundError,MarqoCloudIndexNotReadyError
+
+class TestMarqoUrlResolver(MarqoTestCase):
```
pandu-k: This class should also be renamed

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -0,0 +1,120 @@
...
+            control_base_url="https://api.marqo.ai", api_key="your-api-key", url_cache_duration=60
+        )
+        # Call refresh_urls_if_needed without waiting
+        mapping._refresh_urls_if_needed("index1")
+        initial_timestamp = mapping.latest_index_mappings_refresh_timestamp
+        time.sleep(0.1)
+        mapping._refresh_urls_if_needed("index2")
+
+        # Check that the timestamp has not been updated
+        assert mapping.latest_index_mappings_refresh_timestamp == initial_timestamp
```
pandu-k, line 52: How about also asserting that mock_get was only called once?

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_refresh_urls_if_not_needed(self, mock_get):
+        mock_get.return_value.json.return_value = {"results": [
+            {"index_name": "index1", "endpoint": "example.com", "index_status": "READY"},
+            {"index_name": "index2", "endpoint": "example2.com", "index_status": "READY"}
+        ]}
+        mapping = MarqoCloudInstanceMappings(
+            control_base_url="https://api.marqo.ai", api_key="your-api-key", url_cache_duration=60
+        )
+        # Call refresh_urls_if_needed without waiting
+        mapping._refresh_urls_if_needed("index1")
```
pandu-k, line 45: wouldn't this technically still trigger a refresh? because this happened slightly later than the MarqoCloudInstanceMappings was instantiated? I am concerned that this test may occasionally fail because of this.

We can mock something to prevent the initial refresh, or let the first refresh happen, and then reset the cache to empty and then start doing the actual test

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -0,0 +1,120 @@
...
+        mock_get.return_value.json.return_value = {"results": [
+            {"index_name": "index1", "endpoint": "example.com", "index_status": "READY"},
+            {"index_name": "index2", "endpoint": "example2.com", "index_status": "READY"}
+        ]}
+        mapping = MarqoCloudInstanceMappings(
+            control_base_url="https://api.marqo.ai",api_key="your-api-key", url_cache_duration=60
+        )
+        initial_timestamp = mapping.latest_index_mappings_refresh_timestamp
+
+        # Wait for more than the expiration time
```
pandu-k: This comment needs to be reworded to reflect what is actually happening

tests/marqo_test.py
danyilq: We are running try catch in case index already exists or deleting

src/marqo/cloud_helpers.py
```diff
@@ -0,0 +1,13 @@
+import time
+
+from marqo.marqo_logging import mq_logger
+
+
+def cloud_wait_for_index_status(req, index_name, status):
```
pandu-k: Thanks setting up this file.

1. please add types to function signature
2. please add a docstring. The docstring should describe what it is for and what it does 
3. Please add descriptions for each function parameter. These descriptions should be useful for a new dev to have a high-level understanding of what it does

src/marqo/index.py
```diff
@@ -49,12 +51,27 @@ def __init__(
...
+            try:
+                if self.get_status()["index_status"] != IndexStatus.CREATED:
+                    logging.warning(f"Index {index_name} is not ready. Status: {self.get_status()}, operations may fail.")
+                    skip_version_check = True
+            except Exception as e:
+                skip_version_check = True
+        if not skip_version_check:
+            self._marqo_minimum_supported_version_check()
+        else:
+            logging.warning("Version check is skipped because index is not ready yet.")
```
pandu-k: move this warning about version check to line be a part of the warning on line 58

src/marqo/index.py
```diff
@@ -49,12 +51,27 @@ def __init__(
...
         self.created_at = self._maybe_datetime(created_at)
         self.updated_at = self._maybe_datetime(updated_at)
-        self._marqo_minimum_supported_version_check()
+        skip_version_check = False
+        if config.is_marqo_cloud:
+            try:
+                if self.get_status()["index_status"] != IndexStatus.CREATED:
+                    logging.warning(f"Index {index_name} is not ready. Status: {self.get_status()}, operations may fail.")
+                    skip_version_check = True
+            except Exception as e:
```
pandu-k: 1.  Ideally we figure out the possible exceptions that can occur and handle them intelligently. We should avoid `except Exception`
2. At the very least we need to log exceptions.

src/marqo/index.py
```diff
@@ -49,12 +51,27 @@ def __init__(
         self.index_name = index_name
         self.created_at = self._maybe_datetime(created_at)
         self.updated_at = self._maybe_datetime(updated_at)
-        self._marqo_minimum_supported_version_check()
+        skip_version_check = False
+        if config.is_marqo_cloud:
+            try:
+                if self.get_status()["index_status"] != IndexStatus.CREATED:
+                    logging.warning(f"Index {index_name} is not ready. Status: {self.get_status()}, operations may fail.")
```
pandu-k: Can we provide a more helpful warning?

```
... Common operations, such as search and add_documents, may fail until the index is ready. Please check `mq.index("{index_name}").get_status()` for the index's status. 
```

src/marqo/index.py
```diff
@@ -49,12 +51,27 @@ def __init__(
...
+        else:
+            logging.warning("Version check is skipped because index is not ready yet.")
 
     def delete(self) -> Dict[str, Any]:
         """Delete the index.
         """
-        return self.http.delete(path=f"indexes/{self.index_name}")
+        response = self.http.delete(path=f"indexes/{self.index_name}")
+        if self.config.is_marqo_cloud:
+            cloud_wait_for_index_status(self.http, self.index_name, IndexStatus.DELETED)
```
pandu-k, line 73: The problem is that all the logging in `cloud_wait_for_index_status` assumes that the user is trying to create an index! https://github.com/marqo-ai/py-marqo/blob/e7f514d812024d3b44b52bd97085c684dae7d692/src/marqo/cloud_helpers.py#L11

pandu-k: This is why we need to more clearly define the responsibility of a function (by its name and docstring), so we don't accidentally use functions in unexpected ways (and using tests to guarantee that the behaviour of the function reflects its name/docstring).

tests/marqo_test.py
```diff
@@ -113,10 +115,14 @@ def setUpClass(cls) -> None:
         api_key = os.environ.get("MARQO_API_KEY", None)
         if (api_key):
             local_marqo_settings["api_key"] = api_key
-
+        cls.index_suffix = os.environ.get("MARQO_INDEX_SUFFIX", None)
+        if not cls.index_suffix:
+            os.environ["MARQO_INDEX_SUFFIX"] = str(uuid.uuid4())[:8]
```
pandu-k: nit: I am not sure about the uniqueness guarantee of the first 8 digits of a UUID, but I guess it would be fine for our low-risk use case

tests/marqo_test.py
```diff
@@ -113,10 +115,14 @@ def setUpClass(cls) -> None:
         api_key = os.environ.get("MARQO_API_KEY", None)
         if (api_key):
             local_marqo_settings["api_key"] = api_key
-
+        cls.index_suffix = os.environ.get("MARQO_INDEX_SUFFIX", None)
+        if not cls.index_suffix:
```
pandu-k: This is not thread safe! If we want it to be thread safe we need to move this logic to be execute at the module-level, so it only runs once, when Python first import Marqo 

For example, we could move this to a function, and then call this function at line 25, for example like this:

```
set_marqo_index_suffix()
```

tests/marqo_test.py
```diff
@@ -126,15 +132,34 @@ def tearDownClass(cls) -> None:
         """Delete commonly used test indexes after all tests are run
         """
         client = marqo.Client(**cls.client_settings)
-        commonly_used_ix_name = 'my-test-index-1'
-        indexes_to_tear_down = [cls.generic_test_index_name, commonly_used_ix_name]
-        for ix_name in indexes_to_tear_down:
+        for index in client.get_indexes()['results']:
```
pandu-k, line 163: can we please do validation here? We don't want to accidentally delete the wrong indexes on a dev's machine. We could do something simple like:

```
if index.index_name.startswith('test-index'):
    index.delete()
```

pandu-k: We also need to record all this special logic to a docstring at the top of this file, so devs know the expected behaviour of these tests

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
         for i in range(5):
             func(*args, **kwargs)
 
-    def create_cloud_index(self, index_name, index_defaults=None, **kwargs):
+    def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
+        def create_settings_hash():
```
pandu-k: move nested functions out

tests/marqo_test.py
```diff
@@ -126,15 +132,34 @@ def tearDownClass(cls) -> None:
...
+    def setUp(self) -> None:
+        self.client = Client(**self.client_settings)
+        for index in self.client.get_indexes()['results']:
+            if not self.client.config.is_marqo_cloud:
+                try:
+                    index.delete()
+                except marqo.errors.MarqoApiError as e:
+                    logging.debug(f'received error `{e}` from index deletion request.')
+            else:
+                self.cleanup_documents_from_all_indices()
```
pandu-k: why is this if/else inside the for loop? surely it makes more sense to do:

```
if self.client.config.is_marqo_cloud:
  self.cleanup_documents_from_all_indices()
else:
   for index in self.client.get_indexes()['results']:
       # for loop only triggered if not cloud 
       ... 
```

tests/marqo_test.py
```diff
@@ -126,15 +132,34 @@ def tearDownClass(cls) -> None:
...
+
+    def tearDown(self) -> None:
+        for index in self.client.get_indexes()['results']:
+            if not self.client.config.is_marqo_cloud:
+                try:
+                    index.delete()
+                except marqo.errors.MarqoApiError as e:
+                    logging.debug(f'received error `{e}` from index deletion request.')
+            else:
+                self.cleanup_documents_from_all_indices()
```
pandu-k: ditto on the if/else inside the for loop

tests/marqo_test.py
```diff
@@ -126,15 +132,34 @@ def tearDownClass(cls) -> None:
...
+
+    def tearDown(self) -> None:
+        for index in self.client.get_indexes()['results']:
```
pandu-k: ditto on validation to ensure we only delete test indexes

tests/marqo_test.py
```diff
@@ -126,15 +132,34 @@ def tearDownClass(cls) -> None:
...
             if not client.config.is_marqo_cloud:
                 try:
-                    client.delete_index(ix_name)
+                    index.delete()
                 except marqo.errors.MarqoApiError as e:
                     logging.debug(f'received error `{e}` from index deletion request.')
 
+    def setUp(self) -> None:
+        self.client = Client(**self.client_settings)
+        for index in self.client.get_indexes()['results']:
```
pandu-k: ditto on validation to ensure we only delete test indexes

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
 
-    def create_cloud_index(self, index_name, index_defaults=None, **kwargs):
+    def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
+        def create_settings_hash():
+            combined_dict = {**settings_dict, **kwargs}
+            combined_str = ''.join(f"{key}{value}" for key, value in combined_dict.items())
+            crc32_hash = zlib.crc32(combined_str.encode())
+            short_hash = hex(crc32_hash & 0xffffffff)[2:][
+                         :10]  # Take the first 10 characters of the hexadecimal representation
+            print(f"Created index with settings hash: {short_hash} for settings: {combined_dict}")
```
pandu-k: Thoughts on doing the pattern match on the settings_dict/kwargs to prevent accidentally creating way too many indexes?

pandu-k: The pattern match means that we know ahead of time the maximum number of indexes that can be created

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
+            return short_hash
+
         client = marqo.Client(**self.client_settings)
-        index_settings = index_defaults if index_defaults else {}
-        index_settings.update({
+        settings_dict = settings_dict if settings_dict else {}
+        index_name = f"{index_name}-{self.index_suffix}"
+        if settings_dict or kwargs:
+            index_name = f"{index_name}-{create_settings_hash()}"
```
pandu-k: Not a huge fan of setting vars, when we can just be a lot more explicit about the structure:
```
cloud_test_index_name = f"{index_name}-{self.index_suffix}-{create_settings_hash()}"
```

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
             try:
-                client.create_index(index_name, settings_dict=index_settings)
-            except marqo.errors.MarqoWebError as e:
-                if e.status_code == 409:
-                    continue
-
-    def create_index(self, index_name, index_defaults=None, **kwargs):
+                if client.http.get(f"/indexes/{index_name}/status")["index_status"] == "READY":
+                    break
+            except Exception as e:
```
pandu-k: be more specific than `except Exception as e:` 

The exception may be happening we are operating on an open source index rather than a cloud index

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
-        index_settings = index_defaults if index_defaults else {}
-        index_settings.update({
+        settings_dict = settings_dict if settings_dict else {}
+        index_name = f"{index_name}-{self.index_suffix}"
+        if settings_dict or kwargs:
+            index_name = f"{index_name}-{create_settings_hash()}"
+        settings_dict.update({
             "inference_type": "marqo.CPU", "storage_class": "marqo.basic", "model": "hf/all_datasets_v4_MiniLM-L6"
         })
         while True:
```
pandu-k: Still not 100% clear why we even need to even wait for it to be ready. There is already waiting done when we create a cloud index

src/marqo/index.py
```diff
@@ -141,7 +164,10 @@ def refresh(self):
 
     def get_status(self):
         """gets the status of the index"""
-        return self.http.get(path=F"indexes/{self.index_name}/status")
+        if self.config.is_marqo_cloud:
+            return self.http.get(path=F"indexes/{self.index_name}/status")
+        else:
+            raise UnsupportedOperationError("This operation is only supported for Marqo Cloud")
```
pandu-k, line 173: We need to add a test asserting that this gets error gets thrown for open source

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
-            if client.http.get(f"/indexes/{index_name}/status") == "READY":
-                try:
-                    idx = client.index(index_name)
-                    docs_to_delete = [i['_id'] for i in idx.search("")['hits']]
-                    if docs_to_delete:
-                        idx.delete_documents(docs_to_delete)
-                except marqo.errors.MarqoCloudIndexNotFoundError:
-                    pass
+        for index in indexes['results']:
+            if self.index_suffix in index.index_name.split('-'):
```
pandu-k: validate to ensure only test indexes get deleted (for example checking for `test-index`...)

tests/marqo_test.py
```diff
@@ -144,37 +169,48 @@ def warm_request(self, func, *args, **kwargs):
...
+                if client.http.get(f"/indexes/{index.index_name}/status")["index_status"] == "READY":
+                    docs_to_delete = [i['_id'] for i in index.search("")['hits']]
```
pandu-k: use a large limit to reduce the number of calls needed:
```
index.search("", limit=1000)
```

tests/scripts/delete_all_indexes.py
```diff
@@ -0,0 +1,24 @@
...
+        "url": os.environ.get("MARQO_URL", 'http://localhost:8882'),
+    }
+    api_key = os.environ.get("MARQO_API_KEY", None)
+    if api_key:
+        local_marqo_settings["api_key"] = api_key
+    client = marqo.Client(**local_marqo_settings)
+    indexes = client.get_indexes()
+    for index in indexes['results']:
+        if client.config.is_marqo_cloud:
+            if index.get_status()["index_status"] == marqo.enums.IndexStatus.CREATED:
```
pandu-k: what about other statuses? such as "CREATING" or "PENDING" - we need to be careful around these edge cases

tests/scripts/delete_all_indexes.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+import marqo
+
+
+def cleanup_documents_from_all_indices():
```
pandu-k: inaccurate name.

this should be `delete_all_test_indexes` right?

tests/scripts/delete_all_indexes.py
```diff
@@ -0,0 +1,24 @@
...
+    local_marqo_settings = {
+        "url": os.environ.get("MARQO_URL", 'http://localhost:8882'),
+    }
+    api_key = os.environ.get("MARQO_API_KEY", None)
+    if api_key:
+        local_marqo_settings["api_key"] = api_key
+    client = marqo.Client(**local_marqo_settings)
+    indexes = client.get_indexes()
+    for index in indexes['results']:
```
pandu-k: add validation please to delete only test indexes. This needs to not accidentally delete a random index a dev is using, setting them back a day or so of work

tests/v0_tests/test_add_documents.py
```diff
@@ -59,40 +38,40 @@ def test_create_index_hnsw(self):
...
     # Delete index tests:
 
-    def test_delete_index(self):
-        self.create_index(index_name=self.index_name_1)
-        self.client.delete_index(self.index_name_1)
-        self.create_index(index_name=self.index_name_1)
+    # def test_delete_index(self):
+    #     self.test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
+    #     self.client.delete_index(self.test_index_name)
+    #     self.test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
```
pandu-k: why was this removed? if it's that problematic for the cloud you can just mark it for skipping

tests/v0_tests/test_add_documents.py
```diff
@@ -14,43 +14,22 @@
...
-            except MarqoApiError as s:
-                pass
-        else:
-            self.delete_documents(self.index_name_1)
-
     # Create index tests
-
     def test_create_index(self):
-        self.create_index(index_name=self.index_name_1)
+        self.test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
```
pandu-k: are you trying to set this for all following tests? What guarantee that this runs before any other test runs? Will this work if the tests are ran concurrently?

if you want to ensure this gets set for the entire class before they run why don't you override the `setUpClass` method (ensuring that we're also calling the superclass's  `setUpClass` method) add this to that logic?

tests/marqo_test.py
```diff
@@ -146,4 +167,50 @@ def warm_request(self, func, *args, **kwargs):
...
+        while True:
+            try:
+                if client.http.get(f"/indexes/{index_name}/status")["index_status"] == "READY":
+                    break
+            except Exception as e:
+                pass
+            self.client.create_index(index_name, settings_dict=settings_dict, **kwargs)
+        return index_name
+
+    def create_test_index(self, index_name, settings_dict=None, **kwargs):
```
pandu-k: Make your function signature accurate (include types and return type)

this needs a docs string. Describe what the responsibility of this function is and explain side effects

tests/v0_tests/test_add_documents.py
```diff
@@ -14,43 +14,22 @@
...
+        except MarqoError as e:
+            assert e.code in ["index_already_exists", "index_already_exists_cloud"]
 
     def test_create_index_hnsw(self):
-        if self.client.config.is_marqo_cloud:
-            self.client.delete_index(self.index_name_1)
-        self.create_index(index_name=self.index_name_1, index_defaults={
+        if not self.client.config.is_marqo_cloud:
+            self.client.delete_index(self.generic_test_index_name)
+        self.test_index_name = self.create_test_index(index_name=self.generic_test_index_name, settings_dict={
```
pandu-k: why are you setting `test_index_name` this as an object attribute, when it just used within this test case. 

Ditto for the rest of the occurrences of `self.test_index_name`

tests/v0_tests/test_get_indexes.py
```diff
@@ -44,15 +20,15 @@ def test_get_indexes(self):
         instance.
         """
         ix_0 = self.client.get_indexes()
-        assert not self._is_index_name_in_get_indexes_response(self.index_name_1, ix_0)
+        assert not self._is_index_name_in_get_indexes_response(self.generic_test_index_name, ix_0)
```
pandu-k, line 24: ditto on the weird pattern of reassigning the `self.text_index...` 

we shouldn't be reassigning object attributes unless we need to

tests/v0_tests/test_index.py
```diff
@@ -164,6 +154,7 @@ def run():
             return True
         assert run()
 
+    @mark.ignore_cloud_tests
     def test_create_cloud_index_non_default_param(self):
```
pandu-k, line 165: we should probably test this for the cloud. But if there is an existing non-default-param index we are using for the cloud we make the test against that

tests/v0_tests/test_sentence_chunking.py
```diff
@@ -71,151 +50,135 @@ def test_sentence_chunking_no_overlap(self):
...
-        else:
-            self.delete_documents(self.index_name)
-
         settings = {
             "sentences_per_chunk":2,  
             "sentence_overlap":0 
             }
         
-        self.create_index(self.index_name, **settings)
+        self.test_index_name = self.create_test_index(self.generic_test_index_name, **settings)
```
pandu-k: why are we reassigning `self.test_index_name`??

tox.ini
```diff
@@ -13,3 +13,19 @@ deps =
   numpy
 commands =
   pytest {posargs}
+
+[testenv:cloud_tests]
+passenv =
+    MARQO_CLOUD_URL
+    MARQO_API_KEY
+    MARQO_URL
+    TEST_INDEX_SUFFIX
```
pandu-k: make comments in this file explaining:
- why there is both `MARQO_URL` and `MARQO_CLOUD_URL`
- what `TEST_INDEX_SUFFIX` does and what its structure is 
-  which of these env vars are required and which are optional?

tests/v0_tests/test_telemetry.py
```diff
@@ -62,23 +47,24 @@ def test_telemetry_search(self):
             {"q": "search query","search_method": "LEXICAL"},
             {"q": "search query","searchable_attributes": ["Description"]}]
 
-        self.client.create_index(self.index_name_1)
-        self.client.index(self.index_name_1).add_documents([{"Title": "A dummy document",}], tensor_fields=["Title"])
+        self.test_index_name = self.create_test_index(self.generic_test_index_name)
```
pandu-k: ditto on reassigning `self.test_index_name`

tests/v0_tests/test_tensor_search.py
```diff
@@ -43,41 +31,41 @@ def strip_marqo_fields(doc, strip_id=True):
...
-        search_res = self.client.index(self.index_name_1).search(
+        search_res = self.client.index(self.test_index_name).search(
             "title about some doc")
         assert len(search_res["hits"]) == 0
 
     def test_search_highlights(self):
         """Tests if show_highlights works and if the deprecation behaviour is expected"""
-        self.create_index(index_name=self.index_name_1)
-        self.client.index(index_name=self.index_name_1).add_documents([{"f1": "some doc"}], tensor_fields=["f1"])
+        self.test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
```
pandu-k: reassigning `self.test_index_name`...

src/marqo/index.py
danyilq: @pandu-k there are cases that are unhandled on cloud right now and returning wrong exception right now In this case it just fails index initialisation if Index is not found

tests/scripts/delete_all_indexes.py
danyilq: tests wouldn't finish if some index is creating

danyilq: otherwise we'd just have to wait and something isn't working correct, we could add few minutes waiting time for those that are creating and clear them after that

tests/scripts/delete_all_indexes.py
danyilq: @pandu-k I don't think we ever expect dev to be using account designated for integration tests to run it's own indices

src/marqo/cloud_helpers.py
```diff
@@ -3,11 +3,13 @@
 from marqo.marqo_logging import mq_logger
 
 
-def cloud_wait_for_index_status(req, index_name, status):
-    creation = req.get(f"indexes/{index_name}/status")
-    while creation['index_status'] != status:
+def cloud_wait_for_index_status(req , index_name: str, status):
+    """ Wait for index to be created on Marqo Cloud by checking
+    it's status every 10 seconds until it becomes expected value"""
```
pandu-k: This docstring isn't accurate because `Client.delete` uses this to wait for an index to be deleted. Just reword this docstring to accurately reflect the behaviour and usage of this method

tox.ini
pandu-k: comments: `; this is a comment`

tests/marqo_test.py
```diff
@@ -105,6 +105,16 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+def create_settings_hash(settings_dict, kwargs):
```
pandu-k, line 125: create a doc string about this.
- explain the purpose of settings_hash
- how it fits into the wider context of the cloud test architecture 
- Explain choices, such as why the hash is truncated at 10 chars

tests/marqo_test.py
```diff
@@ -115,10 +125,7 @@ def setUpClass(cls) -> None:
...
         if (api_key):
             local_marqo_settings["api_key"] = api_key
-        cls.index_suffix = os.environ.get("MARQO_INDEX_SUFFIX", None)
-        if not cls.index_suffix:
-            os.environ["MARQO_INDEX_SUFFIX"] = str(uuid.uuid4())[:8]
-            cls.index_suffix = os.environ["MARQO_INDEX_SUFFIX"]
+        cls.index_suffix = os.environ.get("MARQO_INDEX_SUFFIX", "")
         cls.client_settings = local_marqo_settings
         cls.authorized_url = cls.client_settings["url"]
         cls.generic_test_index_name = 'test-index'
```
pandu-k: is `cls.generic_test_index_name` actually an index_name? or is just a prefix? If it is just the prefix, we can rename this var. If it can be both, explain it with a comment

tests/marqo_test.py
pandu-k, line 1: In the docstring for this file can you also describe the `MARQO_INDEX_SUFFIX` env var? And the purpose of this var?

pandu-k: nit: `MARQO_INDEX_SUFFIX` can be more descriptive. For example: `MQ_TEST_RUN_IDENTIFIER` (as it is meant to differentiate tests run indexes from each other)

tests/scripts/create_test_suffix.py
```diff
@@ -0,0 +1,8 @@
+import os
+import uuid
+
+
+def set_index_suffix():
```
pandu-k, line 10: brief doc string describing this func

tests/scripts/delete_all_indexes.py
```diff
@@ -3,22 +3,22 @@
...
     for index in indexes['results']:
-        if client.config.is_marqo_cloud:
-            if index.get_status()["index_status"] == marqo.enums.IndexStatus.CREATED:
-                index.delete()
-        else:
-            index.delete()
+        if index.index_name.startswith('test-index'):
+            if suffix is not None and suffix in index.index_name.split('-'):
+                if index.get_status()["index_status"] == marqo.enums.IndexStatus.CREATED:
+                    index.delete()
```
pandu-k: Question: so we don't delete open source indexes because this is delegated to the marqo_test logic?

If so, then we can rename the file and function to be more specific: `delete_all_cloud_test_indexes.py` for example

tests/scripts/run_cloud_tests.py
```diff
@@ -0,0 +1,18 @@
+import sys
+from create_test_suffix import set_index_suffix
+from delete_all_indexes import delete_all_test_indices
+
+if __name__ == '__main__':
+    # Generate the random suffix
+    set_index_suffix()
+
+    # Run the first command to generate the suffix (already done)
+    # generate_index_suffix.py will set the TEST_INDEX_SUFFIX environment variable
```
pandu-k: is this line comment outdated?

tests/v0_tests/test_client.py
```diff
@@ -52,20 +52,20 @@ def test_error_handling_in_health_check(self):
...
 
     def test_check_index_health_query(self):
+        test_index_name = self.create_test_index(self.generic_test_index_name)
         with patch("marqo._httprequests.HttpRequests.get") as mock_get:
-            self.test_index_name = self.create_test_index(self.generic_test_index_name)
-            res = self.client.index(self.test_index_name).health()
+            res = self.client.index(test_index_name).health()
             args, kwargs = mock_get.call_args
-            self.assertIn(f"/{self.test_index_name}/health", kwargs["path"])
+            self.assertIn(f"health", kwargs["path"])
```
pandu-k, line 56: I actually prefer having the full path (`/{test_index_name}/health`) checked

danyilq: I believe that it returns incorrect response with marqo 1.0.0 but I could be wrong

src/marqo/index.py
```diff
@@ -59,7 +59,7 @@ def __init__(
                                     f"Please check `mq.index('{index_name}').get_status()` for the index's status. "
                                     f"Skipping version check.")
                     skip_version_check = True
-            except Exception as e:
+            except (MarqoWebError, TypeError) as e:
```
pandu-k: do we also need to except KeyError, in case "index_status" is not present in the get_status() return body?

src/marqo/instance_mappings.py
pandu-k: bump on this - feel free to use my suggested sentence as a placeholder until Farshid gets back to you

tests/marqo_test.py
```diff
@@ -1,8 +1,22 @@
...
+- MARQO_API_KEY: The API key used for authentication.
+- MARQO_URL: The URL of the Marqo cloud instance.
+
+When `cluster.is_marqo_cloud` is set to True, the tests will have different setUp and tearDown procedures:
+- Indices will not be deleted after each test.
+- However, documents will be deleted after each test.
+
+The function `create_cloud_index` will be triggered whenever an index needs to be created for the cloud tests.
+It checks the status of the index and creates a new index if needed. It also adds a suffix for unique test execution
+and a hash generated from index settings to ensure each index is unique.
```
pandu-k: Let's make this sentence more accurate:

~~a hash generated from index settings to ensure each index is unique.~~
->
...a hash generated from index settings ensure that only 1 index is created for a given index settings_dict or index kwargs. It also ensures that any tests that require an index with specific settings get routed to the appropriate index.

tox.ini
```diff
@@ -16,10 +16,9 @@ commands =
...
-    MARQO_CLOUD_URL
-    MARQO_API_KEY
-    MARQO_URL
-    TEST_INDEX_SUFFIX
+    MARQO_CLOUD_URL # URL that config uses to resolve whether cluster is cloud v2
+    MARQO_API_KEY # This is the API key used to authenticate with the cloud
+    MARQO_URL # URL that is used as marqo instance url
 whitelist_externals =
   python
 deps =
```
pandu-k, line 24: do we need to add `zlib` somewhere? As it used hashing index settings?

tests/marqo_test.py
```diff
@@ -106,12 +121,16 @@ def wrapper(self, *args, **kwargs):
 
 
 def create_settings_hash(settings_dict, kwargs):
-    combined_dict = {**settings_dict, **kwargs}
-    combined_str = ''.join(f"{key}{value}" for key, value in combined_dict.items())
+    """
+    Creates a hash from the settings dictionary and kwargs. Used to ensure that each index is created unique.
+    Size is restricted on 10 characters to prevent having to big index name which could cause issues.
+    """
+    dict_to_hash = {**settings_dict, **kwargs} if settings_dict else kwargs
```
pandu-k: This may not be sufficient  to  ensure that `{"a": 1, "b":2 }` and `{"b":2, "a": 1}` convert to the same dict.

For example we could hash a serialised dict like here: https://stackoverflow.com/a/22003440

I'd recommend using tried-and-tested methods of serialising dicts than coming up with our own

tests/marqo_test.py
```diff
@@ -180,24 +199,31 @@ def warm_request(self, func, *args, **kwargs):
             func(*args, **kwargs)
 
     def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
+        """
+        Creates a cloud index with the given name and settings. If settings_dict or any kwargs are provided,
+        they will be used to create a unique index name by hashing the settings_dict and kwargs.
+        If settings_dict is not provided, the index name will be the given index_name + the index_suffix.
```
pandu-k: Make sure the language is accurate:

`If settings_dict is not provided,` -> `If settings_dict and index settings kwargs are not provided...`

tests/marqo_test.py
```diff
@@ -180,24 +199,31 @@ def warm_request(self, func, *args, **kwargs):
...
+
+        Performs check for status of index and waits for index to be ready before returning.
+        """
```
pandu-k, line 227: nit:
Explain the function params. Especially that kwargs are those that would be passed to a `Index.create()` call

tests/marqo_test.py
```diff
@@ -180,24 +199,31 @@ def warm_request(self, func, *args, **kwargs):
...
+            })
         try:
-            status = client.http.get(f"/indexes/{index_name}/status")["index_status"]
+            status = client.http.get(f"indexes/{index_name}/status")["index_status"]
             if status == "CREATING":
-                while status == "CREATING":
-                    time.sleep(10)
-                    status = client.http.get(f"/indexes/{index_name}/status")["index_status"]
+                cloud_wait_for_index_status(client.http, index_name, "READY")
             if status != "READY":
```
pandu-k: shouldn't this be `else if`? if status == "CREATING" and then we wait for the index to be created, then the second if statement runs and then the index is attempted to be created again - which probably raises a MarqoWebError, which gets caught in the except block and then the index is attempted to be created a 3rd time.

It is worth re-reviewing this block of code to iron out any odd logic

tests/scripts/create_test_suffix.py
```diff
@@ -1,3 +1,8 @@
+"""
+Creates a unique suffix for the index name to avoid conflicts
+with another testing executed in the same account.
```
pandu-k: nit, language could be more precise:
`testing executed in the same account.` -> `testing executed in the same Marqo cloud account.`

tests/scripts/delete_all_cloud_indexes.py
pandu-k, line 1: the name of this file could still be more precise (for example `delete_all_cloud_test_indexes.py`)

tests/v0_tests/test_add_documents.py
```diff
@@ -51,10 +51,10 @@ def test_create_index_hnsw(self):
...
     # Delete index tests:
 
-    def test_delete_index(self):
-        test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
-        self.client.delete_index(test_index_name)
-        test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
+    # def test_delete_index(self):
+    #     test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
+    #     self.client.delete_index(test_index_name)
+    #     test_index_name = self.create_test_index(index_name=self.generic_test_index_name)
```
pandu-k: why is this commented out? it might still be useful as an open source test

tests/v0_tests/test_client.py
pandu-k: @danyilq this was released in Marqo version 1.1.0 (which is only now being rolled onto the cloud. Please see [here](https://github.com/marqo-ai/marqo/releases/tag/1.1.0)

tests/marqo_test.py
danyilq: in the first part we are checking if index status is CREATING and if so we are entering infinite while loop until index reaches READY, in the end it will just ignore next check if it is READY, else it will trigger creation for that index, it was working fine this way

tests/marqo_test.py
danyilq: now that I think about it, when creating hash there shouldn’t be settings_dict and any kwargs passed simultaneously as settings_dict are prioritised over kwargs anyways. Instead changing it to `dict_to_hash = settings_dict if settings_dict else kwargs`

tox.ini
danyilq: I believe that zlib is a built-in library

tests/v0_tests/test_marqo_cloud_instance_mapping.py
danyilq: The purpose of the test is verifying that 2nd call to index2 won't trigger refresh as mocked response includes both index1 and index2 as READY state, which means that we load both urls in _urls_mapping and timestamp for them is at the first call, 2nd call doesn't trigger a refresh as index2 is loaded in cache and timestamp difference does not trigger cold refresh.

pyproject.toml
```diff
@@ -1,3 +1,6 @@
 [build-system]
 requires = ["setuptools>=42"]
-build-backend = "setuptools.build_meta"
\ No newline at end of file
+build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini_options]
+markers = "ignore_cloud_tests: Mark tests to be ignored when running with env set to cloud_tests"
```
pandu-k: nit: this marker has been confusing me for quite a while.  A better name for it is probably `ignore_during_cloud_tests`

tests/marqo_test.py
```diff
@@ -202,9 +204,16 @@ def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
...
         they will be used to create a unique index name by hashing the settings_dict and kwargs.
-        If settings_dict is not provided, the index name will be the given index_name + the index_suffix.
+        If settings_dict and index settings kwargs are not provided,
+        the index name will be the given index_name + the index_suffix.
 
         Performs check for status of index and waits for index to be ready before returning.
+
+        Args:
+            index_name (str): Name of index to create
+            settings_dict (dict): Dictionary of settings to use when creating index,
```
pandu-k: nit: worth adding a caveat here: `Please note that settings_dict override any kwargs during index creation`

tests/marqo_test.py
pandu-k: yes you are correct - only 1 of <settings_dict> or <kwargs> makes sense.

But you haven't addressed my point. Two settings dicts, with the structure `{"a": 1, "b":2 }` and `{"b":2, "a": 1}` need to resolve to the same string, so that they can get the same hash. this is important to mitigate blowing up the number indexes that end up being made. 

One way we can enforce this is by sorting the keys. I suggest reading that link and using that as your serialisation procedure

tests/marqo_test.py
pandu-k: Yep all good I think I misread your code

pandu-k (dismissed):

pandu-k (dismissed):

farshidz (dismissed):

tests/marqo_test.py
```diff
@@ -226,6 +233,11 @@ def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
             settings_dict.update({
                 "inference_type": "marqo.CPU", "storage_class": "marqo.basic"
             })
+
+        if len(index_name) > 32:
+            index_name = index_name[len(index_name) - 32: len(index_name)]
+            if index_name[0] == "-":
+                index_name = index_name[1:]
```
pandu-k: does this affect any logic that tests for index names starting with a certain pattern?

pandu-k (dismissed):

pandu-k (approved):

PR #134, opened 2023-08-01
Unit test GH workflow

pandu-k:

new CI pipeline

no unit test pipeline for py-marqo

a unit test pipeline for py-marqo

**Test notes**

- Tested and confirmed that this pipeline triggers after commits to a PR branch: https://github.com/marqo-ai/py-marqo/actions/runs/5733088797/ (failing because of a broken unit test)
- Does this pipeline trigger and work after a push to mainline: to be tested
- Does this pipeline work with manually specifying an ECR image: to be tested (must be merged into mainline before testing)
- Does this pipeline work with manually specifying an Docker image: to be tested (must be merged into mainline before testing)

### **Specifications**:

- The test pipeline should run:
  - every time a PR is opened
  - every time code is merged into mainline
  - it can also be run manually
- Marqo should be pulled from an image repo and ran (to keep this pipeline fast)
- When manually running the pipeline, we indicate the Marqo image location, see attached image for a similar existing example
- To keep things simple, Marqo should run OpenSearch in a docker-in-docker configuration

image

#### Determining the Marqo version to use 

The Marqo used for each py-marqo pipeline is:
- Default to the Dockerhub Marqo version specified in py-marqo.supported_marqo_version . 
- Error out if it doesn’t exist

When we run manually we can specify a DockerHub or ECR image tag to run
When testing
- build your feature branch into an ECR image to test against your py-marqo branch


.github/workflows/open-source-unit-tests.yml
```diff
@@ -0,0 +1,136 @@
...
+          aws-access-key-id: ${{ secrets.AWS_EC2_GH_RUNNER_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_EC2_GH_RUNNER_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@v2
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id:  ${{ secrets.AMD_EC2_IMAGE_ID_200GB }}
```
vicilliar, line 53: noting that this means we can't use this for any cuda-related py-marqo tests, but that should be fine, as we don't have any.

.github/workflows/open-source-unit-tests.yml
```diff
@@ -0,0 +1,136 @@
+name: Py-marqo  open source unit tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      push_to:
```
vicilliar: might be a misleading input name, as the image should already be pushed at this time. Maybe `pull_from` or `image_location`

vicilliar (changes requested):
good work on this! this will definitely be super useful.

.github/workflows/open-source-unit-tests.yml
pandu-k: good catch!

pandu-k: Changed to `image_registry_location`

.github/workflows/open-source-unit-tests.yml
pandu-k: Thanks for bringing this up Joshua, it made me consider this use case. 

I think this workflow will stick to CPU-only tests for now. These main set of unit tests should be runnable locally on developers' machines. We can always clone this workflow and change it to create CUDA-dedicated workflow (similar to what we do in the Marqo repo) in the future .

vicilliar (approved):

PR #143, opened 2023-08-10
PyPi-push CI changes: unit tests, include releases/* branches  

pandu-k:

CI pipeline update

PyPi-push CI only runs on pushes to mainline, does not run unit tests.
The unit tests pipeline also runs on pushes to mainline. 

PyPi-push CI now also runs on branches matching the `releases/*` pattern. The pipeline also runs unit tests, which must pass, before being pushed to PyPi.
The unit tests pipeline no longer also runs on pushes to mainline, as these tests are run in the upgraded push-to-pypi workflow
The unit tests pipeline now also runs on update to PRs matching the `releases/*` pattern. 

No

* **Testing**
After changes are merged, the follow tests will be run through:
image

**Notes and observations**
* The GH workflow runs the on-push commands _after_ merging, using the updated workflow.


.github/workflows/open-source-unit-tests.yml
```diff
@@ -20,12 +20,10 @@ on:
       image_tag:
         description: 'Marqo image tag. Examples: "1.1.0", "test" "latest"'
         required: true
-  push:
-    branches: 
-      - mainline
```
pandu-k, line 25: This is removed so that unit tests don't run 3 times for each push to mainline. The other two places the unit tests run, on a push to mainline:
- Once for PR sync action (below)
- Once running the push-to-pypi workflow

.github/workflows/open-source-unit-tests.yml
```diff
@@ -20,12 +20,10 @@ on:
...
   pull_request:
```
vicilliar, line 24: just clarifying this runs unit tests every time there's a push to a mainline or releases branch, right? If so, this might be a good thing to add to the `marqo` repo as well.

.github/workflows/publish-to-PyPI.yml
```diff
@@ -1,18 +1,127 @@
...
+        uses: machulav/ec2-github-runner@v2
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id:  ${{ secrets.AMD_EC2_IMAGE_ID_200GB }}
+          ec2-instance-type: t3.xlarge
+          subnet-id: ${{ secrets.AMD_SUBNET_ID }}
+          security-group-id: ${{ secrets.AMD_SECURITY_GROUP_ID }}
+
+  Test-Py-Marqo:
```
vicilliar, line 46: To lessen the redundant code, this would be a perfect place to use `called workflows`, and call the unit test workflow from here.

.github/workflows/open-source-unit-tests.yml
```diff
@@ -20,12 +20,10 @@ on:
       image_tag:
         description: 'Marqo image tag. Examples: "1.1.0", "test" "latest"'
         required: true
-  push:
```
vicilliar, line 23: you can add `workflow_call:` here so this workflow can be directly called by the publish PyPI workflow

.github/workflows/publish-to-PyPI.yml
```diff
@@ -1,18 +1,127 @@
...
+        run: |
+          docker pull ${{ steps.prepare.outputs.registry }}/${{ steps.prepare.outputs.image_repo }}:${{ steps.prepare.outputs.image_tag }}
+          docker run --name marqo -d --privileged -p 8882:8882 --add-host host.docker.internal:host-gateway \
+            ${{ steps.prepare.outputs.registry }}/${{ steps.prepare.outputs.image_repo }}:${{ steps.prepare.outputs.image_tag }}
+                    
+          # wait for marqo to start with timeout of 10 minutes
+          timeout 10m bash -c 'until [[ $(curl -v --silent --insecure http://localhost:8882 2>&1 | grep Marqo) ]]; do sleep 0.1; done;' || (echo "Marqo did not start in time" && exit 1)
+          
+          python -m pip install --upgrade pip
+          pip install tox
```
vicilliar, line 94: it may generally be a good idea to pin the tox version. i've had issues in the past with deprecation (`whitelist` & `allowlist`)

vicilliar (changes requested):

.github/workflows/open-source-unit-tests.yml
pandu-k: Do you mean the `pull_requests` trigger? It runs every time there is a push to a branch that is in PR directed at mainline/releases. This would be really useful to be added to Marqo too (I plan to add this to Marqo at some point)

.github/workflows/publish-to-PyPI.yml
pandu-k: Good idea. Thanks!

.github/workflows/open-source-unit-tests.yml
pandu-k: Good idea. Added, thanks!

.github/workflows/publish-to-PyPI.yml
pandu-k: Good idea. Pinned.

.github/workflows/open-source-unit-tests.yml
vicilliar: yep i meant the `pull_requests` trigger. Does it also run when the pull request itself is created, or only on pushes to it? Also doing a bit of research, the `pull_request_review` trigger may also be cool to think about. But that can be done another time.

PR #153, opened 2023-08-10
Cloud integration tests enhancements

danyilq:

- Improvement to the speed of cloud integration tests execution
- Makes cloud get_indexes request faster by not sending get_status call for each index initialisation and utlising instance mappings cache for it

* **Other information**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
         "tox"
     ],
     name="marqo",
-    version="1.2.1",
+    version="1.2.2",
     author="marqo org",
     author_email="org@marqo.io",
     description="Tensor search for humans",
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
         "tox"
     ],
     name="marqo",
-    version="1.2.0",
+    version="1.2.1",
     author="marqo org",
     author_email="org@marqo.io",
     description="Tensor search for humans",
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
         "tox"
     ],
     name="marqo",
-    version="1.2.3",
+    version="1.2.4",
     author="marqo org",
     author_email="org@marqo.io",
     description="Tensor search for humans",
--- a/tests/marqo_test.py
+++ b/tests/marqo_test.py
@@ -35,7 +35,7 @@ from marqo.utils import construct_authorized_url
 from marqo._httprequests import HTTP_OPERATIONS
 from marqo.version import __marqo_version__ as py_marqo_support_version
 from marqo.client import Client
-from marqo.errors import InternalError, MarqoApiError, MarqoWebError
+from marqo.errors import InternalError, MarqoApiError, MarqoWebError, MarqoError
 import zlib
 from marqo.cloud_helpers import cloud_wait_for_index_status
 
@@ -122,12 +122,13 @@ def with_documents(index_to_documents_fn: Callable[[], Dict[str, List[Dict[str, 
     return decorator
 
 
-def create_settings_hash(settings_dict, kwargs):
+# replication of create_settings_hash from marqo.marqo_test
+def create_settings_hash(settings_dict, **kwargs):
     """
     Creates a hash from the settings dictionary and kwargs. Used to ensure that each index is created unique.
     Size is restricted on 10 characters to prevent having to big index name which could cause issues.
     """
-    dict_to_hash = settings_dict if settings_dict else kwargs
+    dict_to_hash = settings_dict["index_defaults"] if settings_dict else kwargs
     combined_str = json.dumps(dict_to_hash, sort_keys=True)
     crc32_hash = zlib.crc32(combined_str.encode())
     short_hash = hex(crc32_hash & 0xffffffff)[2:][
@@ -152,6 +153,8 @@ class MarqoTestCase(TestCase):
         cls.generic_test_index_name = 'test-index'  # used as a prefix when index is created with settings
         cls.generic_test_index_name_2 = cls.generic_test_index_name + '-2'
 
+        cls.indexes_to_cleanup = set()
+
         # class property to indicate if test is being run on multi
         cls.IS_MULTI_INSTANCE = (True if os.environ.get("IS_MULTI_INSTANCE", False) in ["True", "TRUE", "true", True] else False)
 
@@ -170,9 +173,7 @@ class MarqoTestCase(TestCase):
 
     def setUp(self) -> None:
         self.client = Client(**self.client_settings)
-        if self.client.config.is_marqo_cloud:
-            self.cleanup_documents_from_all_indices()
-        else:
+        if not self.client.config.is_marqo_cloud:
             for index in self.client.get_indexes()['results']:
                 if index.index_name.startswith(self.generic_test_index_name):
                     try:
@@ -228,7 +229,7 @@ class MarqoTestCase(TestCase):
         client = marqo.Client(**self.client_settings)
         index_name = f"{index_name}-{self.index_suffix}"
         if settings_dict or kwargs:
-            index_name = f"{index_name}-{create_settings_hash(settings_dict, kwargs)}"
+            index_name = f"{index_name}-{create_settings_hash(settings_dict, **kwargs)}"
         if settings_dict:
             settings_dict.update({
                 "inference_type": "marqo.CPU", "storage_class": "marqo.basic"
@@ -246,6 +247,7 @@ class MarqoTestCase(TestCase):
         except (MarqoWebError, TypeError) as e:
             self.client.create_index(index_name, settings_dict=settings_dict,
                                      inference_node_type="marqo.CPU", storage_node_type="marqo.basic", **kwargs)
+        self.indexes_to_cleanup.add(index_name)
         return index_name
 
     def create_test_index(self, index_name: str, settings_dict: dict = None, **kwargs):
@@ -258,13 +260,16 @@ class MarqoTestCase(TestCase):
         return index_name
 
     def cleanup_documents_from_all_indices(self):
-        client = marqo.Client(**self.client_settings)
-        indexes = client.get_indexes()
+        indexes = self.client.get_indexes()
         for index in indexes['results']:
-            if index.index_name.startswith(self.generic_test_index_name) and \
-                    self.index_suffix in index.index_name.split('-'):
-                if client.http.get(f"/indexes/{index.index_name}/status")["index_status"] == "READY":
+            if index.index_name in self.indexes_to_cleanup:
+                print(f"Deleting documents from index {index.index_name}")
+                try:
+                    # veryfying that index is in the mapping
+                    self.client.config.instance_mapping.get_index_base_url(index.index_name)
                     docs_to_delete = [i['_id'] for i in index.search("", limit=100)['hits']]
                     while docs_to_delete:
                         index.delete_documents(docs_to_delete, auto_refresh=True)
                         docs_to_delete = [i['_id'] for i in index.search("", limit=100)['hits']]
+                except MarqoError as e:
+                    print(f"Error deleting documents from index {index.index_name}: {e}")
--- a/tests/marqo_test.py
+++ b/tests/marqo_test.py
@@ -242,10 +242,10 @@ class MarqoTestCase(TestCase):
                 cloud_wait_for_index_status(client.http, index_name, "READY")
             elif status != "READY":
                 self.client.create_index(index_name, settings_dict=settings_dict,
-                                         inference_node_type="marqo.CPU", storage_node_type="marqo.basic", **kwargs)
+                                         inference_node_type="marqo.CPU.large", storage_node_type="marqo.basic", **kwargs)
         except (MarqoWebError, TypeError) as e:
             self.client.create_index(index_name, settings_dict=settings_dict,
-                                     inference_node_type="marqo.CPU", storage_node_type="marqo.basic", **kwargs)
+                                     inference_node_type="marqo.CPU.large", storage_node_type="marqo.basic", **kwargs)
         return index_name
 
     def create_test_index(self, index_name: str, settings_dict: dict = None, **kwargs):
--- a/tests/marqo_test.py
+++ b/tests/marqo_test.py
@@ -130,8 +130,9 @@ def create_settings_hash(settings_dict, **kwargs):
     if settings_dict:
         settings_dict = settings_dict.copy()
         index_defaults = settings_dict.get("index_defaults", {}).copy()
-        del settings_dict["index_defaults"]
-        settings_dict.update(index_defaults)
+        if index_defaults:
+            del settings_dict["index_defaults"]
+            settings_dict.update(index_defaults)
     dict_to_hash = settings_dict if settings_dict else kwargs
     combined_str = json.dumps(dict_to_hash, sort_keys=True)
     crc32_hash = zlib.crc32(combined_str.encode())
@@ -156,8 +157,6 @@ class MarqoTestCase(TestCase):
         cls.authorized_url = cls.client_settings["url"]
         cls.generic_test_index_name = 'test-index'  # used as a prefix when index is created with settings
         cls.generic_test_index_name_2 = cls.generic_test_index_name + '-2'
-
-        cls.indexes_to_cleanup = set()
 
         # class property to indicate if test is being run on multi
         cls.IS_MULTI_INSTANCE = (True if os.environ.get("IS_MULTI_INSTANCE", False) in ["True", "TRUE", "true", True] else False)
@@ -186,9 +185,7 @@ class MarqoTestCase(TestCase):
                         logging.debug(f'received error `{e}` from index deletion request.')
 
     def tearDown(self) -> None:
-        if self.client.config.is_marqo_cloud:
-            self.cleanup_documents_from_all_indices(self.indexes_to_cleanup)
-        else:
+        if not self.client.config.is_marqo_cloud:
             for index in self.client.get_indexes()['results']:
                 if index.index_name.startswith(self.generic_test_index_name):
                     try:
@@ -253,7 +250,7 @@ class MarqoTestCase(TestCase):
         except (MarqoWebError, TypeError) as e:
             self.client.create_index(index_name, settings_dict=settings_dict,
                                      inference_node_type="marqo.CPU.large", storage_node_type="marqo.basic", **kwargs)
-        self.indexes_to_cleanup.add(index_name)
+        self.cleanup_documents_from_index(index_name)
         return index_name
 
     def create_test_index(self, index_name: str, settings_dict: dict = None, **kwargs):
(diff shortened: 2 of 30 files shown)


tests/marqo_test.py
```diff
@@ -228,7 +229,7 @@ def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
         client = marqo.Client(**self.client_settings)
         index_name = f"{index_name}-{self.index_suffix}"
         if settings_dict or kwargs:
```
pandu-k: can we do some validation? if settings_dict AND kwargs: throw error?

tests/marqo_test.py
```diff
@@ -258,13 +260,16 @@ def create_test_index(self, index_name: str, settings_dict: dict = None, **kwarg
         return index_name
 
     def cleanup_documents_from_all_indices(self):
```
pandu-k: Can you add a docstring highlighting this intended for use during cloud tests?

tests/marqo_test.py
```diff
@@ -258,13 +260,16 @@ def create_test_index(self, index_name: str, settings_dict: dict = None, **kwarg
...
-                if client.http.get(f"/indexes/{index.index_name}/status")["index_status"] == "READY":
+            if index.index_name in self.indexes_to_cleanup:
+                print(f"Deleting documents from index {index.index_name}")
+                try:
+                    # veryfying that index is in the mapping
+                    self.client.config.instance_mapping.get_index_base_url(index.index_name)
                     docs_to_delete = [i['_id'] for i in index.search("", limit=100)['hits']]
                     while docs_to_delete:
                         index.delete_documents(docs_to_delete, auto_refresh=True)
                         docs_to_delete = [i['_id'] for i in index.search("", limit=100)['hits']]
```
pandu-k: is it worth doing some error handling? so that we can detect if we're stuck in a potentially infinite loop (like the cloud has a bug preventing us from deleting docs?)

For example:
```
max_loops = 1000
if number_of_while_loops > max_loops:
    raise RuntimeError("Could delete all documents in index {index.index_name} after {max_loops} tries")
```

tests/marqo_test.py
```diff
@@ -152,6 +153,8 @@ def setUpClass(cls) -> None:
         cls.generic_test_index_name = 'test-index'  # used as a prefix when index is created with settings
         cls.generic_test_index_name_2 = cls.generic_test_index_name + '-2'
 
+        cls.indexes_to_cleanup = set()
```
pandu-k: Doesn't that mean there will be one of these dicts per test class? What about using a global dictionary?

tests/scripts/delete_all_cloud_test_indexes.py
```diff
@@ -17,7 +17,7 @@ def delete_all_test_indices():
         if index.index_name.startswith('test-index'):
             if suffix is not None and suffix in index.index_name.split('-'):
                 if index.get_status()["index_status"] == marqo.enums.IndexStatus.CREATED:
-                    index.delete()
+                    index.delete(wait_for_readiness=False)
```
pandu-k: Should we do some waiting until we are confirmed deletion (including retrying + timeout if unsuccessful)? 

We should be careful about not-cleaned-up resources

tests/v0_tests/test_index.py
pandu-k, line 1: more edge cases for health please

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,134 @@
...
+    combined_str = json.dumps(dict_to_hash, sort_keys=True)
+    crc32_hash = zlib.crc32(combined_str.encode())
+    short_hash = hex(crc32_hash & 0xffffffff)[2:][
+                 :10]  # Take the first 10 characters of the hexadecimal representation
+    print(f"Created index with settings hash: {short_hash} for settings: {dict_to_hash}")
+    return short_hash
+
+
+def populate_indices():
+    test_uniqueness_id = os.environ.get("MQ_TEST_RUN_IDENTIFIER", "")
```
pandu-k, line 48: do we need to raise an error if we can't find the target env var?

tests/marqo_test.py
```diff
@@ -122,12 +122,20 @@ def wrapper(self, *args, **kwargs):
...
+# replication of create_settings_hash from marqo.marqo_test
+def create_settings_hash(settings_dict, **kwargs):
     """
     Creates a hash from the settings dictionary and kwargs. Used to ensure that each index is created unique.
     Size is restricted on 10 characters to prevent having to big index name which could cause issues.
     """
-    dict_to_hash = settings_dict if settings_dict else kwargs
+    if settings_dict:
+        index_defaults = settings_dict.get("index_defaults", {}).copy()
+        if index_defaults and len(settings_dict) != 1:
```
pandu-k: why check ` if ... settings_defaults != 1`?

pandu-k: And why do we delete and re-add it?

tests/marqo_test.py
```diff
@@ -122,12 +122,20 @@ def wrapper(self, *args, **kwargs):
...
+            del settings_dict["index_defaults"]
+            settings_dict.update(index_defaults)
+    print(settings_dict)
+    quit()
```
pandu-k: i am guessing this is just for testing. ..

tests/marqo_test.py
pandu-k: And how it cleans up the indexes in `indexes_to_cleanup`

Also, I prefer if `indexes_to_cleanup` was given to this function as an argument

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+        "cool-index" + '-' + test_uniqueness_id:
+            [{}],
+        "first-index" + '-' + test_uniqueness_id:
+            [{}],
+        "weight-index" + '-' + test_uniqueness_id:
+            [{}],
+        "mmodal-index" + '-' + test_uniqueness_id:
+            [{"treat_urls_and_pointers_as_images": True, "model": "ViT-B/32"}],
+        generic_test_index_name + '-' + test_uniqueness_id:
+            [{},
```
pandu-k, line 54: Can you add a docstring/comments about why a single index name maps to multiple index settings ?

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+    api_key = os.environ.get("MARQO_API_KEY", None)
+    if api_key:
+        marqo_settings["api_key"] = api_key
+
+    mq = marqo.Client(**marqo_settings)
+
+    indexes_to_create = []
+
+    for index_name, configs_list in index_name_to_config_mappings.items():
+        for config in configs_list:
```
pandu-k: calling this config is confusing (I thought it was referring to an instance of the Config class). How about calling this index_configs, for example?

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+                     "ann_parameters": {
+                         "parameters": {
+                             "m": 24
+                         }
+                     }
+                 }
+             }
+              },
+             {"treat_urls_and_pointers_as_images": True,
+              "model": "ViT-B/32",
```
pandu-k, line 66: We definitely need to add to docstring/comments that if settings_dict is not included in the dicts, then it implies that the dict represents the index creation kwargs

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+    mq = marqo.Client(**marqo_settings)
+
+    indexes_to_create = []
+
+    for index_name, configs_list in index_name_to_config_mappings.items():
+        for config in configs_list:
+            settings_dict = config.get("settings_dict")
+            if config:
+                if not settings_dict:
+                    index_name_to_create = index_name + '-' + create_settings_hash(settings_dict=None, **config)
```
pandu-k: is it possible to do `create_settings_hash(settings_dict=None, kwargs=**config)` for readability?

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+    return short_hash
+
+
+def populate_indices():
+    test_uniqueness_id = os.environ.get("MQ_TEST_RUN_IDENTIFIER", "")
+    generic_test_index_name = 'test-index'
+    index_name_to_config_mappings = {
+        generic_test_index_name + '-2' + '-' + test_uniqueness_id:
+            [{}],
+        "cool-index" + '-' + test_uniqueness_id:
```
pandu-k: why don't we have `create_cloud_index_name_base("cool-index", test_uniqueness_id)`?

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+    for index_name, configs_list in index_name_to_config_mappings.items():
+        for config in configs_list:
+            settings_dict = config.get("settings_dict")
+            if config:
+                if not settings_dict:
+                    index_name_to_create = index_name + '-' + create_settings_hash(settings_dict=None, **config)
+                else:
+                    index_name_to_create = index_name + '-' + create_settings_hash(settings_dict=settings_dict)
+            else:
+                index_name_to_create = index_name
```
pandu-k, line 134: Does this mean for those 4 demo indexes, it creates 4 different indexes? If their settings are the same they should be able to share the same cloud index, unless one test requires 2 different indexes to be set up at the same time,

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+        print(f"Creating {index_name} with config: {config}")
+        print(mq.create_index(index_name=index_name, wait_for_readiness=False, **config))
+
+    while True:
+        if all(creating_index in mq.config.instance_mapping._urls_mapping["READY"] .keys()
+               for creating_index, _ in indexes_to_create):
+            break
+        mq.config.instance_mapping._refresh_urls()
+        time.sleep(10)
+        print("Waiting for indexes to be created...")
```
pandu-k, line 163: Probably need to have some way to timeout and raise a runtime error

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -0,0 +1,142 @@
...
+    dict_to_hash = settings_dict["index_defaults"] if settings_dict else kwargs
+    combined_str = json.dumps(dict_to_hash, sort_keys=True)
+    crc32_hash = zlib.crc32(combined_str.encode())
+    short_hash = hex(crc32_hash & 0xffffffff)[2:][
+                 :10]  # Take the first 10 characters of the hexadecimal representation
+    print(f"Created index with settings hash: {short_hash} for settings: {dict_to_hash}")
+    return short_hash
+
+
+def populate_indices():
```
pandu-k, line 38: shouldn't the mapping and, related functionality, of 
```
index_name <> settings <> cloud_testing_index_name 
```
Be available and easy to use for the entire test dir (rather than separating this logic into this file)?

tests/v0_tests/test_index.py
pandu-k, line 1: more edge cases for the health endpoint

pandu-k, line 1: Also, are we missing tests for any other new methods?

tests/marqo_test.py
danyilq: @pandu-k that's exactly what we want, purpose of this field is to reduce the amount of calls to delete_documents api. so with making it that way will make it so that class cleans up only those indices that were used during it's own tests.

tests/scripts/delete_all_cloud_test_indexes.py
danyilq: @pandu-k if something goes wrong with deleting indexes it's already the same as tests failing, I think we should have warning for the integration tests account if some instance attached to it is running for more than 2 hours or it's up to us to decide.

tests/scripts/populate_indices_for_cloud_tests.py
danyilq: @pandu-k yes, it could be correct, when I was applying changes I seen that we don't use generic name there and use 4 different names instead so I wasn't sure if it is important to keep them on completely fresh index and doc deletion might not help.

tests/marqo_test.py
```diff
@@ -122,20 +122,18 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
-# replication of create_settings_hash from marqo.marqo_test
 def create_settings_hash(settings_dict, **kwargs):
     """
     Creates a hash from the settings dictionary and kwargs. Used to ensure that each index is created unique.
     Size is restricted on 10 characters to prevent having to big index name which could cause issues.
     """
```
pandu-k: ... we de-nest the index_defaults so that some kwarg indexes and setttings_dict indexes are more likely to generate the same index name, saving us from creating an unnecessary index:
```
settings_dict (original) = {"index_defaults": {"a": 1}, "number_of_shards": 2}
-> settings_dict (which we will use for hashing) = {"a": 1, "number_of_shards": 2}
# which matches the following index kwargs:
kwargs = {"a": 1, "number_of_shards": 2}
```

tests/marqo_test.py
```diff
@@ -122,20 +122,18 @@ def wrapper(self, *args, **kwargs):
...
     if settings_dict:
+        settings_dict = settings_dict.copy()
```
pandu-k: settings_dict_simplified = ... .copy()

tests/marqo_test.py
```diff
@@ -266,17 +262,24 @@ def create_test_index(self, index_name: str, settings_dict: dict = None, **kwarg
...
+        idx = self.client.index(index_to_cleanup)
+        max_attempts = 1000
+        print(f"Deleting documents from index {idx.index_name}")
+        try:
+            # veryfying that index is in the mapping
+            idx.refresh()
+            self.client.config.instance_mapping.get_index_base_url(idx.index_name)
+            docs_to_delete = [i['_id'] for i in idx.search("", limit=100)['hits']]
+            attempt = 0
+            while docs_to_delete:
```
pandu-k: Also consider: `index.get_stats()["documentCount"] > 0 ...`

pandu-k: perhaps add refresh to loop

tests/scripts/delete_all_cloud_test_indexes.py
pandu-k:
```
if (after waiting for an 20 min for all indexes to be deleted):
    raise RunTimeErrror(f"problem deleting indexes. please manually clean up indexes: {index-1, index-2}")

```

tests/scripts/populate_indices_for_cloud_tests.py
```diff
@@ -6,12 +6,19 @@
 import marqo
 
 
+# replication of create_settings_hash from marqo.marqo_test
```
pandu-k, line 9: try to get the importing to work to remove duplicate code

tests/marqo_test.py
```diff
@@ -235,11 +229,13 @@ def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
         """
```
pandu-k, line 210: We should throw an error if the cloud index doesn't yet exist

tests/marqo_test.py
pandu-k, line 1: Add to docstring: "We will not actually create and delete real cloud indexes during this test suite, because this slows down py-marqo <> Marqo cloud tests. We should still test these methods with mocking. End-to-end creation and deletion tests will be done elsewhere."

tests/marqo_test.py
pandu-k: Make sure the  error message is helpful for new devs: they should know where to register their new index and index settings.

tests/scripts/run_cloud_tests.py
```diff
@@ -1,19 +1,36 @@
...
+    signal.signal(signal.SIGINT, handle_interrupt)
 
-    import pytest
-    pytest_args = ['tests/', '-m', 'not ignore_during_cloud_tests'] + sys.argv[1:]
-    pytest.main(pytest_args)
-    print("All tests has been executed, proceeding to delete indices")
+    test_start_time = time.time()
+    try:
+        set_unique_run_identifier()
+        print(f"Using unique identifier: {os.environ.get('MQ_TEST_RUN_IDENTIFIER', '')}")
```
pandu-k: let's just fail if the env var is not set: `os.environ["MQ_TEST_RUN_IDENTIFIER"]`

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -110,6 +110,17 @@ def test_request_of_creating_index_raises_error(self, mock_get):
         with self.assertRaises(MarqoCloudIndexNotReadyError):
             mapping.get_index_base_url("index2")
 
+    @patch("requests.get")
+    def test_modifying_state_returns_as_ready(self, mock_get):
```
pandu-k, line 145: do a test for logging (make sure we aren't getting warnings logged)

tests/v0_tests/test_index.py
pandu-k, line 1: Add test for get_indexes() unable to update cache: we should log a warning but continue

tests/marqo_test.py
```diff
@@ -241,11 +226,12 @@ def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
...
             elif status != "READY":
-                self.client.create_index(index_name, settings_dict=settings_dict,
-                                         inference_node_type="marqo.CPU.large", storage_node_type="marqo.basic", **kwargs)
+                raise RuntimeError(f"Indexes must be precreated during cloud tests. Index {index_name} is not ready. "
+                                   f"Please specify it's settings in scripts/populate_all_indices.py")
         except (MarqoWebError, TypeError) as e:
-            self.client.create_index(index_name, settings_dict=settings_dict,
-                                     inference_node_type="marqo.CPU.large", storage_node_type="marqo.basic", **kwargs)
+            raise RuntimeError(f"Indexes must be precreated during cloud tests. Index {index_name} is not ready. "
+                               f"Please specify it's settings in scripts/populate_all_indices.py")
```
pandu-k: nit, grammar: `its` not `it's`

setup.py
```diff
@@ -18,7 +18,7 @@
         "tox"
     ],
     name="marqo",
-    version="1.2.2",
+    version="1.2.3",
```
pandu-k: this needs to be bumped again. See here: https://pypi.org/project/marqo/

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -134,6 +134,175 @@ def test_request_of_not_existing_index_raises_error(self, mock_get):
         with self.assertRaises(MarqoCloudIndexNotFoundError):
             mapping.get_index_base_url("index2")
 
+    @patch("requests.get")
+    def test_get_indexes_fails_cache_doesnt_update(self, mock_get):
```
pandu-k, line 167: this is a good test. What about a similar test, but the first time trying to `get_indexes` fails?

pandu-k (dismissed):

PR #158, opened 2023-08-14
Remove time based cloud index URL cache refresh

farshidz:

Performance improvement

Marqo Cloud index URL cache refreshes every 6 minutes

Cache will only refresh if we try to resolve an index for which there is no URL in cache. Furthermore, Marqo client evicts index cached URL if there's a `ConnectionError` when trying to reach the data plane. This means the next request on that index will trigger a cache refresh.

The following behaviour is expected:
* If the user deletes an index after it's been cached (e.g., searched), the first request on that index will return `BackendCommunicationError`. Any further attempts to reach the deleted index will return `MarqoCloudIndexNotFoundError`.
* If the user deletes an index after it's been cached, and creates a new index with the same name, the first attempt to access the new index with the same client instance will return a `BackendCommunicationError`. Further attempts will resolve to the new index correctly.
* Arbitrary network issues on a valid index will lead to cache eviction and refresh on the next request (but refresh frequency is capped at one per 15 seconds regardless).

No

* **Other information**:

### DIFF
--- a/src/marqo/instance_mappings.py
+++ b/src/marqo/instance_mappings.py
@@ -37,9 +37,9 @@ class InstanceMappings(ABC):
         pass
 
     @abstractmethod
-    def on_instance_error(self, index_name: str, http_status: Optional[int] = None) -> None:
+    def index_http_error_handler(self, index_name: str, http_status: Optional[int] = None) -> None:
         """
-        Called when an HTTP error occurs on a Marqo instance.
+        Called when an HTTP error occurs on a Marqo index operation.
 
         Args:
             index_name: The index name
--- a/src/marqo/instance_mappings.py
+++ b/src/marqo/instance_mappings.py
@@ -37,7 +37,7 @@ class InstanceMappings(ABC):
         pass
 
     @abstractmethod
-    def on_instance_error(self, index_name: str, http_status: int) -> None:
+    def on_instance_error(self, index_name: str, http_status: Optional[int] = None) -> None:
         """
         Called when an HTTP error occurs on a Marqo instance.
 
--- a/src/marqo/instance_mappings.py
+++ b/src/marqo/instance_mappings.py
@@ -35,3 +35,14 @@ class InstanceMappings(ABC):
     @abstractmethod
     def is_remote(self):
         pass
+
+    @abstractmethod
+    def on_instance_error(self, index_name: str, http_status: int) -> None:
+        """
+        Called when an HTTP error occurs on a Marqo instance.
+
+        Args:
+            index_name: The index name
+            http_status: The HTTP status code
+        """
+        pass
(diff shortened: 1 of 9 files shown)


src/marqo/instance_mappings.py
```diff
@@ -35,3 +35,14 @@ def get_control_base_url(self) -> str:
     @abstractmethod
     def is_remote(self):
         pass
+
+    @abstractmethod
+    def on_instance_error(self, index_name: str, http_status: int) -> None:
```
pandu-k: as in on a specific Marqo dataplane? We can probably call this `index_http_error_handler` to be more descriptive. Is it OK to use "index" over "instance" - I think the abstractions still hold.

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -68,3 +65,9 @@ def _refresh_urls(self, timeout=None):
                 self._urls_mapping[IndexStatus.CREATING][index['index_name']] = index.get('endpoint')
         if self._urls_mapping:
             self.latest_index_mappings_refresh_timestamp = time.time()
+
+    def index_http_error_handler(self, index_name: str, http_status: Optional[int] = None) -> None:
```
pandu-k, line 69: Thoughts on this solution?
```
def index_http_error_handler(self, index_name: str, http_status: Optional[int] = None) -> None:
    # the logic here prevents refreshing faster than 15s anyway:  
    _refresh_urls_if_needed(index_name=index_name) 
    
    # when control returns Marqo's HTTPRequests module, it raises the original error
    return
...
```

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -65,9 +66,7 @@ def _refresh_urls(self, timeout=None):
...
 
-    def on_instance_error(self, index_name: str, http_status: int) -> None:
-        if http_status == 404:
-            mq_logger.debug(f'Evicting index {index_name} from cache (if exists) due to 404')
+    def index_http_error_handler(self, index_name: str, http_status: Optional[int] = None) -> None:
+        mq_logger.debug(f'Refreshing cache due to error {http_status} on index {index_name}')
 
-            self._urls_mapping['READY'].pop(index_name, None)
-            self._urls_mapping['CREATING'].pop(index_name, None)
+        self._refresh_urls()
```
pandu-k: Sorry I mixed up the methods. This should be `_refresh_urls_if_needed()` (which has the logic around url_cache_duration)

pandu-k: Sorry I mixed up the methods. This should be `_refresh_urls_if_needed()` (which has the logic around url_cache_duration)

src/marqo/marqo_cloud_instance_mappings.py
farshidz: Addressed

src/marqo/marqo_cloud_instance_mappings.py
farshidz: Duplicate

pandu-k (dismissed):

pandu-k (approved):

PR #160, opened 2023-08-14
Reduce warnings on index instantiation

pandu-k:
Bug fix. 

There are too many unnecessary warning messages while setting up an index.

Reduce the number of warning messages on index creation

No 

### DIFF
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -82,6 +82,7 @@ class Client:
             inference_node_count=1,
             storage_node_count=1,
             replicas_count=0,
+            wait_for_readiness=True
     ) -> Dict[str, Any]:
         """Create the index. Please refer to the marqo cloud to see options for inference and storage node types.
 
@@ -101,6 +102,7 @@ class Client:
             inference_node_count;
             storage_node_count:
             replicas_count:
+            wait_for_readiness:
         Returns:
             Response body, containing information about index creation result
         """
@@ -112,21 +114,25 @@ class Client:
             image_preprocessing_method=image_preprocessing_method,
             settings_dict=settings_dict, inference_node_type=inference_node_type, storage_node_type=storage_node_type,
             storage_node_count=storage_node_count, replicas_count=replicas_count,
-            inference_node_count=inference_node_count,
+            inference_node_count=inference_node_count, wait_for_readiness=wait_for_readiness
         )
 
-    def delete_index(self, index_name: str) -> Dict[str, Any]:
+    def delete_index(self, index_name: str, wait_for_readiness=True) -> Dict[str, Any]:
         """Deletes an index
 
         Args:
             index_name: name of the index
-
+            wait_for_readiness: Marqo Cloud specific, whether to wait until
+                operation is completed or to proceed without waiting for status,
+                won't do anything if config.is_marqo_cloud=False
         Returns:
             response body about the result of the delete request
         """
         try:
             res = self.http.delete(path=f"indexes/{index_name}")
-            cloud_wait_for_index_status(self.http, index_name, enums.IndexStatus.DELETED)
+            if self.config.is_marqo_cloud and wait_for_readiness:
+                cloud_wait_for_index_status(self.http, index_name, enums.IndexStatus.DELETED)
+                return res
         except errors.MarqoWebError as e:
             return e.message
 
--- a/src/marqo/client.py
+++ b/src/marqo/client.py
@@ -132,7 +132,7 @@ class Client:
             res = self.http.delete(path=f"indexes/{index_name}")
             if self.config.is_marqo_cloud and wait_for_readiness:
                 cloud_wait_for_index_status(self.http, index_name, enums.IndexStatus.DELETED)
-                return res
+            return res
         except errors.MarqoWebError as e:
             return e.message
 
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -613,6 +613,7 @@ class Index:
                 f"If you are sure your Marqo version is compatible with this client, you can ignore this message. ")
 
             if url not in marqo_url_and_version_cache:
+                # self.get_marqo() uses get_index_base_url(), so it should be available
                 marqo_url_and_version_cache[url] = self.get_marqo()["version"]
 
             marqo_version = marqo_url_and_version_cache[url]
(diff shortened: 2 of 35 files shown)


src/marqo/index.py
```diff
@@ -52,19 +52,28 @@ def __init__(
...
         try:
             self.config.instance_mapping.get_index_base_url(self.index_name)
         except errors.MarqoError as e:
-            mq_logger.warning(e)
+            mq_logger.debug(
+                f'Cache update on index object instantiation could not retrieve index URL: {e}')
             skip_version_check = True
-        if not skip_version_check:
+
+        if (self.config.instance_mapping.is_index_usage_allowed(index_name=self.index_name)
```
danyilq, line 63: isn't `is_index_usage_allowed` serving the same purpose the same as `self.config.instance_mapping.get_index_base_url(self.index_name)` above on line 57?

pandu-k: This is because now `get_index_base_url` doesn't care whether an index is CREATING or READY. `is_index_usage_allowed`, on the other hand, does care.

src/marqo/client.py
```diff
@@ -112,21 +114,25 @@ def create_index(
...
+                won't do anything if config.is_marqo_cloud=False
         Returns:
             response body about the result of the delete request
         """
         try:
             res = self.http.delete(path=f"indexes/{index_name}")
-            cloud_wait_for_index_status(self.http, index_name, enums.IndexStatus.DELETED)
+            if self.config.is_marqo_cloud and wait_for_readiness:
+                cloud_wait_for_index_status(self.http, index_name, enums.IndexStatus.DELETED)
+                return res
```
farshidz: No return for non-cloud? Or indentation is wrong?

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -25,21 +26,27 @@ def get_control_base_url(self) -> str:
...
+        """
         self._refresh_urls_if_needed(index_name)
-        if index_name in self._urls_mapping[IndexStatus.READY]:
-            return self._urls_mapping[IndexStatus.READY][index_name]
-        if index_name in self._urls_mapping[IndexStatus.CREATING]:
-            raise MarqoCloudIndexNotReadyError(index_name)
+
+        for cloud_status, indexes in self._urls_mapping.items():
+            if index_name in indexes:
+                return indexes[index_name]
```
farshidz, line 38: This is going to miss `MarqoCloudIndexNotReadyError` (index creating). Also that our tests aren't catching this

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -25,21 +26,27 @@ def get_control_base_url(self) -> str:
...
 
-    def _refresh_urls_if_needed(self, index_name):
-        if index_name not in self._urls_mapping[IndexStatus.READY] and \
-                time.time() - self.latest_index_mappings_refresh_timestamp > self.url_cache_duration:
-            # fast refresh to catch if index was created
-            self._refresh_urls()
+    def _refresh_urls_if_needed(self, index_name: Optional[str] = None):
+        if index_name is None or index_name not in self._urls_mapping[IndexStatus.READY]:
+            if time.time() - self.latest_index_mappings_refresh_timestamp > self.url_cache_duration:
+                # fast refresh to catch if index was created
```
farshidz: Comment doesn't apply anymore. Not a fast refresh

pandu-k:
Cloud tests pass, too: 
image

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -56,6 +63,7 @@ def _refresh_urls(self, timeout=None):
 
         if not response.ok:
             mq_logger.warning(response.text)
+            return None
```
farshidz, line 65: Since this is something we missed before, did we add the missing test?

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -67,9 +75,20 @@ def _refresh_urls(self, timeout=None):
...
+    def is_index_usage_allowed(self, index_name: str) -> bool:
+        """Checks the status of the index in self._urls_mapping.
+
+        Note that this method does not request a refresh of the mappings, so the result
+        of this method is a best effort.
+
+        If the index_name cannot be found in the self._urls_mapping, then False will be returned.
+        """
+
+        if index_name in self._urls_mapping[IndexStatus.READY]:
```
farshidz, line 90: Since we don't have periodical refreshes anymore, if a user relied on this and never tried any other operation on the index (which will trigger a cache refresh), they would get False indefinitely

src/marqo/client.py
pandu-k: actioned

src/marqo/marqo_cloud_instance_mappings.py
pandu-k: This method's contract has shifted in this PR. The idea is to simplify the `get_index_base_url()` logic. We let users' context that an index is creating (from the console, for example) to interpret any 500 error they receive when they prematurely try to use the index.

pandu-k: The relevant test is here: https://github.com/marqo-ai/py-marqo/blob/pandu/no-warnings-on-index-instantiation/tests/v0_tests/test_marqo_cloud_instance_mapping.py#L174

src/marqo/marqo_cloud_instance_mappings.py
pandu-k: actioned

src/marqo/marqo_cloud_instance_mappings.py
pandu-k: Good point. Added relevant test: https://github.com/marqo-ai/py-marqo/blob/pandu/no-warnings-on-index-instantiation/tests/v0_tests/test_marqo_cloud_instance_mapping.py#L174

src/marqo/marqo_cloud_instance_mappings.py
pandu-k: Yep. I mention that this is a best effort. This is intended to used to reflect the state of whether the client thinks it's OK to start using the index, rather than attempting to reflect the reality of the index state. 

Currently it is just used to help determine whether to skip an index's version check.  For the default mappings, it always returns True, as the client should always be allowed to check the index's version.

Let me know if you can think of a better abstraction. Perhaps we should just make it more specific and not attempt to be general `is_index_version_check_allowed()`?

farshidz (dismissed):

farshidz (approved):

PR #165, opened 2023-08-17
improve cloud tests setUp/tearDown

danyilq:

* **Other information**:


tests/cloud_tests/run_cloud_tests.py
```diff
@@ -0,0 +1,59 @@
...
+    sys.exit(1)
+
+
+def convert_string_to_boolean(string_value):
+    valid_representations_of_true = ['true', '1']
+    valid_representations_of_false = ['false', '0']
+    if string_value.lower() in valid_representations_of_true:
+        return True
+    elif string_value.lower() in valid_representations_of_false:
+        return False
```
pandu-k: no final else. We should not allow ambiguous input. Raise an error

tests/marqo_test.py
```diff
@@ -112,7 +113,10 @@ def wrapper(self, *args, **kwargs):
             for index_name, docs in index_to_documents.items():
                 if len(docs) == 0:
                     continue
-                self.create_test_index(index_name=index_name)
+                self.create_test_index(
```
pandu-k, line 163: specify this behaviour (the type of index that will be set up) in the docstring

tests/marqo_test.py
```diff
@@ -258,13 +281,17 @@ def cleanup_documents_from_index(self, index_to_cleanup: str):
             self.client.config.instance_mapping.get_index_base_url(idx.index_name)
             attempt = 0
             q = ""
-            while idx.get_stats()["numberOfDocuments"] > 0:
+            docs_to_delete = [i['_id'] for i in idx.search(q, limit=100)['hits']]
+            verified_run = 0
+            while idx.get_stats()["numberOfDocuments"] > 0 or docs_to_delete or verified_run < 3:
```
pandu-k: should we put a short sleep somwhere in this loop (like 0.5 seconds, for example)?

tests/v0_tests/test_image_chunking.py
```diff
@@ -22,12 +22,16 @@ def test_image_no_chunking(self):
                 pass
 
         settings = {
```
pandu-k: nit: specify these are open source settings

.github/workflows/cloud-integration-test.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Run cleanup_indexes.py
+        env:
+          MARQO_URL: ${{ secrets.CLOUD_MARQO_URL }}
+          MARQO_CLOUD_URL: ${{ secrets.CLOUD_MARQO_URL }}
+          MARQO_API_KEY: ${{ secrets.CLOUD_MARQO_API_KEY }}
+        run: python tests/cloud_tests/cleanup_indexes.py
```
pandu-k: this file doesn't exist

.github/workflows/cloud-integration-test.yml
```diff
@@ -0,0 +1,73 @@
+name: Cloud Integration Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      job_to_run:
+        description: 'Job to run'
```
pandu-k: because this is a literal, please clearly state all the possible options a user can input

.github/workflows/cloud-integration-test.yml
pandu-k, line 1: Add a comment to the top of this file describing its behaviour

.github/workflows/cloud-integration-test.yml
pandu-k: And the behaviour of each input

.github/workflows/cloud-integration-test.yml
pandu-k: Also, what the side effects are in each situation

.github/workflows/cloud-integration-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        required: true
+        default: 'cleanup_indexes'
+  pull_request:
```
pandu-k: Let's make this run on any PR review, (not on each PR update)

.github/workflows/cloud-integration-test.yml
```diff
@@ -0,0 +1,73 @@
+name: Cloud Integration Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      job_to_run:
+        description: 'Job to run'
+        required: true
+        default: 'cleanup_indexes'
```
pandu-k: 1. set the default to not cleanup_indexes
2. We consider cleanup_indexes to be an emergency measure. We should rename it and document it (perhaps to `delete_all_test_indexes` appropriately. 
3. This should delete _all_ indexes with the `test-index` prefix

src/marqo/marqo_cloud_instance_mappings.py
```diff
@@ -16,7 +16,7 @@
 class MarqoCloudInstanceMappings(InstanceMappings):
 
     def __init__(self, control_base_url, api_key=None, url_cache_duration: int = 15):
-        self.latest_index_mappings_refresh_timestamp = time.time() - url_cache_duration
+        self.latest_index_mappings_refresh_timestamp = time.time() - url_cache_duration - 1
```
pandu-k, line 19: why this change?

danskiyq (approved):

src/marqo/marqo_cloud_instance_mappings.py
danyilq: Some of the tests were failing due to the too fast execution of steps between instance_mapping initialisation and call for urls refresh. Whilst it is unlikely to happen to any user this other second makes it 100% that on the first call to some index refresh will be triggered.

tests/marqo_test.py
danyilq: with the new way of docs cleanup it should ideally not even trigger, and this was working fine by far so I think it is unnecessary

tests/marqo_test.py
danyilq: could you please elaborate

.github/workflows/cloud-integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      job_to_run:
+        description: 'Job to run (Options: delete_all_indexes, run_integration_tests)'
+        required: true
+        default: 'delete_all_indexes'
+  pull_request:
+    branches:  # temporarily not deleting to simplify testing process
```
pandu-k: ok cool. so this pull_request block will be deleted before merging?

.github/workflows/cloud-integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+#          REPO_NAME="py-marqo"
+#
+#          # Check if the reviewer has write access
+#          write_access=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+#            "https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/collaborators/$REVIEWER_USERNAME/permission" | jq -r '.permission == "write"')
+#
+#          if [[ "$write_access" == "true" ]]; then
+#            echo "Reviewer has write access."
+#          else
+#            echo "Reviewer does not have write access or did not approve the PR."
```
pandu-k: I don't think we need this logic. the `environment: cloud-tests` statement means that only registered people can approve a deployment of this pipeline, regardless of who approved the PR

.github/workflows/cloud-integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+
+name: Cloud Integration Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      job_to_run:
+        description: 'Job to run (Options: delete_all_indexes, run_integration_tests)'
+        required: true
+        default: 'delete_all_indexes'
```
pandu-k: the default should be `run_integration_tests`, as `delete_all_indexes` is meant to be an emergency option

.github/workflows/cloud-integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+# Keep in mind that delete_all_indexes will delete ALL indexes in the
+# integration tests account that starts with the prefix "test-index".
+
+name: Cloud Integration Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      job_to_run:
+        description: 'Job to run (Options: delete_all_indexes, run_integration_tests)'
```
pandu-k: We need to clarify what here what the impact of the `delete_all_indexes` option is. For example, that it deletes _all_ test indexes on the cloud account, even those that may have been started by other users' tests.

tests/cloud_tests/delete_all_cloud_test_indexes.py
```diff
@@ -0,0 +1,46 @@
+import os
+import requests
+import marqo
+
+
+def delete_all_test_indices(wait_for_readiness=False):
```
pandu-k, line 6: Add docstring explaining behaviour and side effects. Basically anything that changes besides the output of the function, and any inputs that aren't explicitly parameters to the function.

So in this case, side effects include (but are not limited to):
- MQ_TEST_RUN_IDENTIFIER, which isn't a parameter to the function, affects the behaviour of this function. Explain its impact
- The deletion of indexes on a cloud index. Explain what indexes are deleted, and how the specific cloud account is specified (i.e. by env vars)

tests/cloud_tests/run_cloud_tests.py
```diff
@@ -0,0 +1,60 @@
...
+tests_specific_kwargs = {
+    'create-indexes': False, 'delete-indexes': False, 'use-unique-identifier': False,
+}
+
+
+def handle_interrupt(signum, frame):
+    print("\nInterrupt received. Cleaning up and deleting indices.")
+    if tests_specific_kwargs['delete-indexes']:
+        delete_all_test_indices(wait_for_readiness=False)
+    sys.exit(1)
```
pandu-k: shouldn't we exit with the same signum, rather than hardcoding 1? What are your thoughts?

tests/cloud_tests/run_cloud_tests.py
```diff
@@ -0,0 +1,60 @@
+import os
+import signal
+import sys
+
+from create_and_set_cloud_unique_run_identifier import set_unique_run_identifier
+from delete_all_cloud_test_indexes import delete_all_test_indices
+from populate_indices_for_cloud_tests import populate_indices
+
+tests_specific_kwargs = {
```
pandu-k, line 17: Create a comment explaining what this dict is for

tests/cloud_tests/run_cloud_tests.py
pandu-k, line 1: Add a docstring explaining usage, including options and syntax

pandu-k: Like explain the meaning of the `use-unique-identifier` option

tests/cloud_tests/run_cloud_tests.py
```diff
@@ -0,0 +1,60 @@
...
+        if 'MQ_TEST_RUN_IDENTIFIER' not in os.environ:
+            os.environ['MQ_TEST_RUN_IDENTIFIER'] = 'cinteg'
+        print(f"Using unique identifier: {os.environ['MQ_TEST_RUN_IDENTIFIER']}")
+        if tests_specific_kwargs['create-indexes']:
+            populate_indices()
+        print(f"All indices has been created, proceeding to run tests with pytest. Arguments: {sys.argv[1:]}")
+
+        import pytest
+        pytest_args = ['tests/', '-m', 'not ignore_during_cloud_tests'] + sys.argv[1:]
+        print(pytest_args)
```
pandu-k: Add context so that a user understand what the args are:

`running cloud integration test with args: {pytest_args}`

tests/cloud_tests/run_cloud_tests.py
```diff
@@ -0,0 +1,60 @@
...
+        pytest_exit_code = pytest.main(pytest_args)
+        if pytest_exit_code != 0:
+            raise RuntimeError(f"Pytest failed with exit code: {pytest_exit_code}")
+        print("All tests has been executed, proceeding to delete indices")
```
pandu-k: `proceeding to delete indices` should only print if `tests_specific_kwargs['delete-indexes']` right?

tests/marqo_test.py
pandu-k: Sorry. Please specify in this function's docstring that its behaviour is to create indexes

tests/marqo_test.py
```diff
@@ -185,87 +195,99 @@ def warm_request(self, func, *args, **kwargs):
...
-    def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
+    def create_test_index(
+            self,
+            cloud_test_index_to_use: CloudTestIndex, open_source_test_index_name: str,
+            open_source_index_settings: dict = None, open_source_index_kwargs: dict = None,
+    ):
+        """Infers whether test is being run on cloud or open source and
+        creates the appropriate index if it being run on open source,
+        otherwise returns the name of the cloud index to use and applies
+        unique run identifier to the index name and performs documents cleanup.
```
pandu-k: Run on sentence. Break it up into 2 or 3 simple sentences

pandu-k: Also, doesn't it always return the name of the index (regardless if it is open source of cloud)? We just need to ensure our language is precise and accurate

tests/marqo_test.py
```diff
@@ -185,87 +195,99 @@ def warm_request(self, func, *args, **kwargs):
...
+        else:
+            index_name_to_return = self.create_open_source_index(
+                open_source_test_index_name, open_source_index_settings, open_source_index_kwargs
+            )
+        return index_name_to_return
 
-            Caveats:
-            - Please note that settings_dict overrides any kwargs during index creation.
-        """
+    def prepare_cloud_index_for_test(self, index_name: str):
```
pandu-k, line 276: Docstring needed. Make sure you include all side effects

pandu-k: A side effect includes dealing with `add_documents_and_mark_for_cleanup_patch`

tests/marqo_test.py
pandu-k, line 1: We need a docstring under `MarqoTestCase` explaining it's behaviour including the new `add_documents_and_mark_for_cleanup_patch` attribute

tests/marqo_test.py
```diff
@@ -185,87 +195,99 @@ def warm_request(self, func, *args, **kwargs):
...
-                cloud_wait_for_index_status(client.http, index_name, "READY")
-            elif status != "READY":
-                raise RuntimeError(f"Indexes must be precreated during cloud tests. Index {index_name} is not ready. "
-                                   f"Please specify it's settings in scripts/populate_all_indices.py")
-        except (MarqoWebError, TypeError) as e:
-            raise RuntimeError(f"Indexes must be precreated during cloud tests. Index {index_name} is not ready. "
-                               f"Please specify its settings in scripts/populate_all_indices.py")
-        self.cleanup_documents_from_index(index_name)
+            raise ValueError("Only one of settings_dict and kwargs can be specified")
+        kwargs = kwargs or {}
```
pandu-k: not a huge fan of mutating parameters. I prefer this pattern:
```
index_creation_settings = {"index_name": index_name}

# you might want to double check logic
if kwargs:
    index_creation_settings.update(**kwargs)
else:
    index_creation_settings(settings_dict=settings_dict)

client.create_index(**index_creation_settings)
```

tests/marqo_test.py
```diff
@@ -185,87 +195,99 @@ def warm_request(self, func, *args, **kwargs):
...
+        # print out all active mocks
+        for active_mock in mock._patch._active_patches:
+            unclosed_mocks[active_mock.kwargs['side_effect']] = active_mock.kwargs['side_effect']([], get_ix_name=True)
+        print(f"unclosed mocks: {unclosed_mocks}")
+        self.add_documents_and_mark_for_cleanup_patch.stop()
+        res = self.client.index(index_name).add_documents(documents, *args, **kwargs)
+        self.add_documents_and_mark_for_cleanup_patch.start()
+        res_list = [res] if type(res) is not list else res
+        for result_to_process in res_list:
+            if not result_to_process['errors']:
```
pandu-k, line 324: you need to handle the case when there is one document that couldn't be indexed because of an error but some other docs _were_ successfully indexed

.github/workflows/cloud-integration-tests.yml
danyilq: yes

tests/marqo_test.py
danyilq: at least in the current test suite there is not a single occurence of such case, and I couldn't reproduce I don't think we need to account for such case as we still have the default cleanup that uses search which would occur if numberOfDocs is not 0 after this one.

tests/marqo_test.py
```diff
@@ -10,13 +10,12 @@
...
 
-The function `create_cloud_index` will be triggered whenever an index needs to be created for the cloud tests.
-It checks the status of the index and creates a new index if needed. It also adds a suffix for unique test execution
-and a hash generated from index settings ensure that only 1 index is created for a given
-index settings_dict or index kwargs.
-It also ensures that any tests that require an index with specific settings get routed to the appropriate index.
+The function `prepare_cloud_index_for_test` will be triggered whenever an index is called for the cloud tests.
+It performs cleanup for index documents and mocks the 'add_documents' method to call
+add_documents_and_mark_for_cleanup_patch instead which stores list of added documents,
+for future cleanup of documents.
```
pandu-k: run-on sentence. Add a full stop between after `instead`, then start a new sentence.

tests/marqo_test.py
```diff
@@ -107,6 +106,8 @@ def side_effect(http_operation, path, body=None, content_type=None, index_name="
 
 
 def with_documents(index_to_documents_fn: Callable[[], Dict[str, List[Dict[str, Any]]]], warmup_query: Optional[str] = None):
+    """Function decorator to create indices and add documents to them.
+    Created index has default settings and CloudTestIndex.basic_index is used during cloud tests."""
```
pandu-k: Wording:
`Created index has default settings and CloudTestIndex.basic_index is used during cloud tests.` 
Is easier to read as:
`The created index has default settings. For cloud tests, CloudTestIndex.basic_index is used.`

tests/marqo_test.py
```diff
@@ -221,6 +226,20 @@ def create_test_index(
         return index_name_to_return
 
     def prepare_cloud_index_for_test(self, index_name: str):
+        """
+        Cleans up documents from the specified cloud index and prepares it for testing.
+
+        This method first cleans up any existing documents in the provided 'index_name'.
+        It then ensures that the 'add_documents' method of the Marqo index is correctly mocked
+        to perform it with proper handling by adding _ids of documents
+        to the self.index_to_documents_cleanup_mapping.
```
pandu-k, line 283: Add why the the _ids are added: `... for later deletion during <RELEVANT FUNCTIONS>`

tests/marqo_test.py
```diff
@@ -221,6 +226,20 @@ def create_test_index(
...
+
+        Also checks for existing patch and stops it if it exists.
```
pandu-k: Explain why it does this.

tests/marqo_test.py
```diff
@@ -236,8 +255,12 @@ def create_open_source_index(self, index_name: str, settings_dict: dict = None,
         client = marqo.Client(**self.client_settings)
         if settings_dict and kwargs:
```
pandu-k: What if settings_dict is an empty dict? Isn't this valid settings_dict value? 

Can we be more explicit to prevent against related edge cases? ?

tests/marqo_test.py
```diff
@@ -266,7 +289,10 @@ def cleanup_documents_from_index(self, index_to_cleanup: str):
             self.client.config.instance_mapping.get_index_base_url(idx.index_name)
 
             if self.index_to_documents_cleanup_mapping.get(index_to_cleanup):
-                idx.delete_documents(list(self.index_to_documents_cleanup_mapping[index_to_cleanup]), auto_refresh=True)
+                res = idx.delete_documents(list(self.index_to_documents_cleanup_mapping[index_to_cleanup]),
+                                           auto_refresh=True)
+                if res['status'] == 'succeeded':
+                    self.index_to_documents_cleanup_mapping[index_to_cleanup] = set()
```
pandu-k, line 347: Are you sure it is safe to do this? Isn't it safer to iterate through each result in `res`, removing successfully deleted docs from `documents_to_cleanup`?

tests/v0_tests/test_add_documents.py
```diff
@@ -33,15 +43,19 @@ def test_create_index_double(self):
     def test_create_index_hnsw(self):
         if not self.client.config.is_marqo_cloud:
             self.client.delete_index(self.generic_test_index_name)
-        test_index_name = self.create_test_index(index_name=self.generic_test_index_name, settings_dict={
-            "index_defaults": {
-                "ann_parameters": {
-                    "parameters": {
-                        "m": 24
+        test_index_name = test_index_name = self.create_test_index(
```
pandu-k: ???

tests/v0_tests/test_add_documents.py
```diff
@@ -551,6 +609,30 @@ def test_add_docs_logs_deprecation_warning_if_non_tensor_fields(self):
...
         with self.assertLogs('marqo', level='WARNING') as cm:
-            test_index_name = self.create_test_index(self.generic_test_index_name)
+            test_index_name = self.create_test_index(
+                cloud_test_index_to_use=CloudTestIndex.basic_index,
+                open_source_test_index_name=self.generic_test_index_name,
+            )
             self.client.index(test_index_name).add_documents(documents=documents, non_tensor_fields=non_tensor_fields)
             self.assertTrue({'`non_tensor_fields`', 'Marqo', '2.0.0.'}.issubset(set(cm.output[0].split(" "))))
+
+    def test_add_documents_mock_overwrites_mock_for_cloud_tests(self):
```
pandu-k: Can you please add a docstring for this test?

pandu-k: What about checking if regular mock activities work?

E.g.: does `mock_add_documents.assert_called_once` work?

pandu-k: Also, if this is a test specifically to guarantee that the add_documents mocking structure works (which is a really smart idea to have!) let's move it into a new test file. For example, `test_cloud_integration_tests.py`

tests/v0_tests/test_index.py
```diff
@@ -209,7 +219,11 @@ def test_create_custom_number_of_replicas(self):
         settings = {
             "number_of_replicas": intended_replicas,
         }
-        test_index_name = self.create_test_index(index_name=self.generic_test_index_name, settings_dict=settings)
+        test_index_name = self.create_test_index(
+            cloud_test_index_to_use=CloudTestIndex.basic_index,
```
pandu-k: if this test is ignored, can you not set the cloud_test_index_to_use? As it doesn't make sense here.

tests/marqo_test.py
```diff
@@ -185,87 +196,117 @@ def warm_request(self, func, *args, **kwargs):
         for i in range(5):
             func(*args, **kwargs)
 
-    def create_cloud_index(self, index_name, settings_dict=None, **kwargs):
-        """
-            Create a cloud index with the given name and settings.
+    def create_test_index(
+            self,
+            cloud_test_index_to_use: CloudTestIndex, open_source_test_index_name: str,
+            open_source_index_settings: dict = None, open_source_index_kwargs: dict = None,
```
pandu-k: is it possible for `open_source_index_settings` to be renamed to `open_source_index_settings_dict`, just to be 100% clear how this parameter is used? 

Or at least clearly explain this parameter (and other args while you're at it) in the docstring)

tests/marqo_test.py
danyilq: out of 200 executions it didn't fail at all, anyways we have default deletion by search after it so it should be safe!

tests/marqo_test.py
pandu-k, line 1: Can you make sure that, in the docstring of `mark_for_cleanup_and_add_documents` that you highlight that this method acts like a decorator, allowing the add_documents calls to propagate through will storing the successfully added doc _ids for later deletion. 

I just realised that this method is a key part of this refactor's logic, and so we need to make it clear what is happening.

tests/v0_tests/test_cloud_integration_tests.py
```diff
@@ -0,0 +1,30 @@
+import os
+from unittest import mock
+
+from tests.marqo_test import MarqoTestCase, CloudTestIndex
+
+
+class TestCloudIntegrationTests(MarqoTestCase):
+    def test_add_documents_mock_overwrites_mock_for_cloud_tests(self):
```
pandu-k, line 10: nice. thanks for the docstring. How about if I create a `mock_add_documents`, where I actually allow documents to be sent to the index, do they get deleted as expected?

For example:
```

 # pseudocode 

def my_test(self, mock_documents):

    mock_documents.side_effect = lambda *args, **kwargs: {
        **add_documents(*args, **kwargs), "my_custom_response_key": 123
    }
    assert "hello" not in self. add_documents_and_mark_for_cleanup_patch
    with mock.patch("marqo.index.Index.add_documents", mock_documents):
         test_add_docs_res = ix.add_documents({"_id": "hello", "blah": "blah"})

    # check if we have correctly marked the ID for deletion:
    assert "hello" in self. index_to_documents_cleanup_mapping[ix.index_name]
    
    # check if we our test's mock worked:
    assert "my_custom_response_key" in test_add_docs_res

    # check if the document did end up in the index  
    assert ix.get_documents_by_id(["hello"]) is as expected

```

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -517,25 +517,6 @@ def test_transitioning_flow_without_modifying(self, mock_get):
         with self.assertRaises(MarqoCloudIndexNotFoundError):
             mapping.get_index_base_url("index1")
 
-    def test_index_instantiation_does_not_refresh_urls(self):
```
pandu-k, line 511: 1. Why was this deleted?
2. Docstring needed. For example `This test ensures that when an index object is instantiated, it doesn't cause a refresh of the instance_mapping's urls_mappings, using latest_index_mappings_refresh_timestamp as a proxy indicator`

pandu-k (commented):
test review

pandu-k (commented):
test rev

pandu-k:
Confirmed that testing locally works:

pandu-k:
Confirmed that testing locally works:
image

danyilq (commented):
trigger workflow

danyilq (commented):
test

src/marqo/enums.py
```diff
@@ -16,3 +16,4 @@ class IndexStatus:
     DELETED = "DELETED"
     MODIFYING = "MODIFYING"
     CREATING = "CREATING"
+    DELETING = "DELETING"
```
pandu-k: What is the possible impact of this? I am nervous this could introduce a regression...

tests/cloud_tests/delete_all_cloud_test_indexes.py
```diff
@@ -27,10 +29,18 @@ def delete_all_test_indices(wait_for_readiness=False):
...
+        return
+
+    print("Indices to delete: ", indices_to_delete)
+    print("Marqo Cloud deletion responses:")
     for index_name in indices_to_delete:
         index = client.index(index_name)
         if index.get_status()["index_status"] == marqo.enums.IndexStatus.READY:
-            index.delete(wait_for_readiness=False)
+            print(index_name, index.delete(wait_for_readiness=False))
+        elif index.get_status()["index_status"] == marqo.enums.IndexStatus.DELETING:
```
pandu-k: Rather than introducing a new index state can you do a string comparison? If we need this new enum we can introduce it in a new PR

tests/v0_tests/test_cloud_integration_tests.py
```diff
@@ -27,4 +29,27 @@ def test_add_documents_mock_overwrites_mock_for_cloud_tests(self):
         if marqo_url_before_test:
             self.client_settings["url"] = marqo_url_before_test
         if marqo_cloud_url_before_test:
-            os.environ["MARQO_CLOUD_URL"] = marqo_cloud_url_before_test
\ No newline at end of file
+            os.environ["MARQO_CLOUD_URL"] = marqo_cloud_url_before_test
+
+    def test_index_cleanup_works_with_add_documents_mocked(self):
```
pandu-k, line 34: add docstring

pandu-k: Few questions:
1. what is this test trying to accomplish? 
2. Was this test I recommended?
3. If this was an attempt at the test I recommended, why does this one deviate?

danyilq: The only deviation is that instead of mocking it with side effect of adding docs I just stop mock on the moment of adding documents and then start it again and it tests that cleanup of documents worked anyways.

danyilq (commented):
test

tests/v0_tests/test_cloud_integration_tests.py
```diff
@@ -34,22 +34,35 @@ def test_add_documents_mock_overwrites_mock_for_cloud_tests(self):
     def test_index_cleanup_works_with_add_documents_mocked(self):
```
pandu-k, line 34: docstring needed. it should explain what this test is trying to test

tests/v0_tests/test_cloud_integration_tests.py
```diff
@@ -34,22 +34,35 @@ def test_add_documents_mock_overwrites_mock_for_cloud_tests(self):
...
+            if self.index_to_documents_cleanup_mapping.get(test_index_name) is not None:
+                assert "lost" not in self.index_to_documents_cleanup_mapping.get(test_index_name)
+
+            add_documents_response = self.client.index(test_index_name).add_documents(
+                documents=[{"some": "data", "_id": "lost"}], tensor_fields=["some"]
+            )
+
+            assert add_documents_response["specific-test-key"] == 123
+            assert "lost" in self.index_to_documents_cleanup_mapping.get(test_index_name)
+            assert self.client.index(test_index_name).get_stats()["numberOfDocuments"] == 1
```
pandu-k, line 67: we need to also ensure that this document actually ended up being indexed properly. get_stats() isn't sufficient.

For example:
```
assert self.client.index(test_index_name).get_document_by_id('lost')['some'] == 'data'
```

danyilq (commented):
trigger cloud tests

pandu-k (approved):

PR #167, opened 2023-08-24
implement bulk search for cloud and add tests coverage for it

danyilq:

* **Other information**:


danyilq (commented):
trigger cloud tests

src/marqo/client.py
```diff
@@ -278,3 +281,33 @@ def raise_error_for_cloud(function_name: str = None):
...
+        """
+        cluster = None
+        index_names = set([q.index for q in parsed_queries])
+        for index_name in index_names:
+            self.index(index_name)  # it will perform all basic checks for index readiness
+            if cluster is None:
+                cluster = self.config.instance_mapping.get_index_base_url(index_name)
+            if cluster != self.config.instance_mapping.get_index_base_url(index_name):
+                raise errors.InvalidArgError(
+                    "All indexes in bulk search request must belong to the same cluster."
```
pandu-k: Let's make this make more sense for for cloud users:
```
All indexes in a bulk search request must belong to the same Marqo cluster. 
- If you are using Marqo Cloud, make sure all search requests in your bulk search use the sane index
```

tests/marqo_test.py
```diff
@@ -128,6 +129,9 @@ def side_effect(http_operation, path, body=None, content_type=None, index_name="
                             if isinstance(response, InternalError):
                                 raise response
                             return response
+                        # performs check for get indexes which needs to pass for mappings during cloud tests
+                        if http_operation == "get" and path == "" and body is None and content_type is None:
+                            return
```
pandu-k: Why special logic for this? don't we need to do things like track call count?

tests/v0_tests/test_get_settings.py
```diff
@@ -27,7 +27,6 @@ def test_default_settings(self):
         self.assertIn("number_of_replicas", index_settings)
         self.assertTrue(fields.issubset(set(index_settings['index_defaults'])))
 
-    @mark.ignore_during_cloud_tests
```
pandu-k, line 30: I'd prefer this to be ignored for cloud tests, because we aren't actually creating a new cloud index

pandu-k: What we can test instead, is to mock HTTP and then assert that the create_index request has the correct body and parameters. This will work for testing cloud.

tests/v0_tests/test_bulk_search.py
```diff
@@ -106,6 +112,7 @@ def test_bulk_search_with_context(self):
             }
         )
     ], forbid_extra_calls=True)
```
pandu-k, line 120: Add a line between methods

tests/v0_tests/test_bulk_search.py
```diff
@@ -60,6 +65,7 @@ def strip_marqo_fields(doc, strip_id=True):
...
```
pandu-k, line 68: newline between class mthods

tests/marqo_test.py
```diff
@@ -176,6 +180,75 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+class InstanceMappingIndexData(NamedTuple):
```
pandu-k: Can you add a docstring describing:
- What this is? (is this meant to mock some existing py-marqo functionality, or is to help better structure the py-marqo test suite logic?)
- What it is intended to do?

tests/marqo_test.py
```diff
@@ -176,6 +180,75 @@ def wrapper(self, *args, **kwargs):
...
+    This decorator is used to mock the behavior of the requests.get function
+    within a specific test function. It allows you to set up mock responses for
+    requests to the "/indexes" URL while allowing real requests to other URLs.
+
+    Args:
+        indexes_list (Union[List[InstanceMappingIndexData], None]): A list of
+            InstanceMappingIndexData objects to be used for mocking responses
+            to requests to the "/indexes" URL. If None, no mock data will be used.
+        to_return_mock (bool): Indicates whether the decorated test function
+            should return the mock object or not. If True, the test function
```
pandu-k: Describe what scenarios the py-marqo test developer would want return the  mocked object, and what scenarios they wouldn't

tests/marqo_test.py
```diff
@@ -176,6 +180,75 @@ def wrapper(self, *args, **kwargs):
...
+    @staticmethod
+    def get_basic_index_data():
+        return InstanceMappingIndexData(
+            index_name="test-index",
+            index_status="READY",
+            endpoint="endpoint"
+        )
+
+
+def mock_instance_mappings(indexes_list: Union[List[InstanceMappingIndexData], None], to_return_mock: bool = False):
```
pandu-k: this function name is not specific enough. 

`mock_get_indexes_http_requests`

tests/marqo_test.py
```diff
@@ -176,6 +180,75 @@ def wrapper(self, *args, **kwargs):
...
+    """Function decorator to mock the instance mappings endpoint.
+
+    This decorator is used to mock the behavior of the requests.get function
+    within a specific test function. It allows you to set up mock responses for
+    requests to the "/indexes" URL while allowing real requests to other URLs.
+
```
pandu-k: add a short message explaining to onboarding devs why `requests.get` was mocked rather than `mq.get_indexes()`

tests/marqo_test.py
pandu-k: Also this class name is confusing. let's call it something that really reflects what this is meant to do. Maybe MockGetIndexesResponseIndexObject

tests/marqo_test.py
```diff
@@ -176,6 +180,75 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+class InstanceMappingIndexData(NamedTuple):
+    index_name: str
+    index_status: str
+    endpoint: str
+
+    @staticmethod
+    def get_basic_index_data():
```
pandu-k: This is meant to be a helper method to get a commonly used InstanceMappingIndexData configuration right? 

I prefer to keep classes as simple as possible. If you want to modularise this then I recommend creating a new file.

For example `tests/cloud_test_logic/models/instance_index_mappings.py`. Then you can import it.

(yep I also think we should rename `cloud_tests` to `cloud_tests_logic`)

Also I think this function should be renamed to make it more clear

tests/marqo_test.py
```diff
@@ -248,32 +321,48 @@ def create_test_index(
             self,
             cloud_test_index_to_use: Union[CloudTestIndex, None], open_source_test_index_name: Union[str, None],
             open_source_index_settings_dict: dict = None, open_source_index_kwargs: dict = None,
+            delete_index_documents_before_test: bool = True
```
pandu-k, line 250: why do we have to adjust this function's parameters?

I'd prefer keeping this simple

tests/marqo_test.py
```diff
@@ -248,32 +321,48 @@ def create_test_index(
...
+            ValueError: If 'cloud_test_index_to_use' is None in cloud environments.
+
         """
         client = marqo.Client(**self.client_settings)
         if client.config.is_marqo_cloud:
             if cloud_test_index_to_use is None:
                 raise ValueError("cloud_test_index_to_use must be specified for cloud tests")
             index_name_to_return = f"{cloud_test_index_to_use.value}-{self.index_suffix}"
-            self.prepare_cloud_index_for_test(index_name_to_return)
+            self.prepare_cloud_index_for_test(index_name_to_return, delete_index_documents_before_test)
```
pandu-k, line 289: shouldn't this be always True?

pandu-k: why would we need documents between tests??

tests/v0_tests/test_bulk_search.py
pandu-k: Why do we need to mock the get_indexes response?

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -170,58 +158,47 @@ def test_refresh_urls_graceful_timeout_handling_http_timeout(self, mock_get):
...
+    def test_request_of_not_existing_index_raises_error(self):
         mapping = MarqoCloudInstanceMappings(
             control_base_url="https://api.marqo.ai", api_key="your-api-key", url_cache_duration=60
         )
         with self.assertRaises(MarqoCloudIndexNotFoundError):
             mapping.get_index_base_url("index2")
 
-    @patch("requests.get")
+    @mock_instance_mappings([InstanceMappingIndexData("index1", IndexStatus.READY, "example.com"),
+                             InstanceMappingIndexData("index2", IndexStatus.MODIFYING, "example2.com")],
```
pandu-k: I would prefer some of these tests to use the string (`"MODIFYING"`) rather than an enum, because this is what is returned by requests

tests/v0_tests/test_bulk_search.py
pandu-k, line 1: I suggest 2 more tests:
- Ensure we can bulk search across 2 different indexes in open source 
- Ensure we cannot bulk search across 2 different indexes in cloud

tests/marqo_test.py
danyilq: if we don't add specific catch that triggers only on requests.get sent by mappings as far as I am concerned the next stroke with forbid_extra_calls would raise errors

tests/v0_tests/test_get_settings.py
danyilq: I am not sure, isn't purpose of the test in getting index settings and verifying that they match? That's possible without mocking http there as it is not checked for open source as well and get_indexes is being tested there.
Or do you mean that we could change index settings and not be aware that it would fail this test?

tests/v0_tests/test_bulk_search.py
danyilq: arent they both decorators for function? it would violate E304

tests/v0_tests/test_bulk_search.py
danyilq: We need to mock it for cloud test scenario as vaildation for bulk uses get_index_endpoint to resolve the cluster it is pointing to and if we don't mock it'd fail during cloud tests

danyilq: as for newline, same question as above, it violates E304

tests/marqo_test.py
danyilq: it is rather not between tests but in some cases e.g. with_documents decorator uses create_test_index and then adds documents and after that following function is calling create_test_index too to get index name to point to and if we do cleanup then documents that were created in with_documents decorator will be simply deleted.

tests/marqo_test.py
danyilq: described reason below

tests/v0_tests/test_marqo_cloud_instance_mapping.py
danyilq: some? so do it 50/50? our Enum represents that string anyways right?

danyilq (commented):
run cloud tests

danyilq (commented):
trigger cloud tests

.github/workflows/cloud-integration-tests.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           MARQO_URL: ${{ secrets.CLOUD_MARQO_URL }}
           MARQO_CLOUD_URL: ${{ secrets.CLOUD_MARQO_URL }}
           MARQO_API_KEY: ${{ secrets.CLOUD_MARQO_API_KEY }}
-        run: tox -e cloud_tests -- create-indexes=True use-unique-identifier=True delete-indexes=True
+        run: tox -e cloud_test_logic -- create-indexes=True use-unique-identifier=True delete-indexes=True
```
pandu-k: this should still be `cloud_tests`. Is this coupled to the directory name? If not can we please reverse it?

danyilq (commented):
test run

src/marqo/client.py
```diff
@@ -307,7 +307,9 @@ def _validate_all_indexes_belong_to_the_same_cluster(self, parsed_queries: List[
                 cluster = self.config.instance_mapping.get_index_base_url(index_name)
             if cluster != self.config.instance_mapping.get_index_base_url(index_name):
                 raise errors.InvalidArgError(
-                    "All indexes in bulk search request must belong to the same cluster."
+                    "All indexes in a bulk search request must belong to the same Marqo cluster."
+                    "- If you are using Marqo Cloud, make sure all search requests"
```
pandu-k, line 311: Do you need a newline?

tests/cloud_test_logic/cloud_instance_mappings.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def mock_get_indexes_response(indexes_list: Union[List[GetIndexesIndexResponseObject], None], to_return_mock: bool = False):
+    """Function decorator to mock the get indexes endpoint.
+
+    This decorator is used to mock the behavior of the requests.get which is used by
+    MarqoCloudInstanceMappings object to retrieve and store specific information about
+    cloud indexes: index_name, index status and index endpoint.
+    It allows you to set up mock responses for requests to the "/indexes" URL
+    while allowing real requests to other URLs.
```
pandu-k: nice! I understand the intention when I read it!

tests/cloud_test_logic/cloud_instance_mappings.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        indexes_list (Union[List[GetIndexesIndexResponseObject], None]): A list of
+            InstanceMappingIndexData objects to be used for mocking responses
+            to requests to the "/indexes" URL. If None, no mock data will be used.
```
pandu-k: Do you mean `if None` then `get_indexes` will return a `get_indexes` response that contains no index information?

Remember to think from the intended audience's perspective

tests/cloud_test_logic/cloud_instance_mappings.py
```diff
@@ -0,0 +1,96 @@
...
+        to_return_mock (bool): Indicates whether the decorated test function
+            should return the mock object or not. If True, the test function
```
pandu-k: what mock object? The mock get object? Or the mock get_indexes response?

tests/marqo_test.py
```diff
@@ -13,7 +13,7 @@
 
 # RUNNING CLOUD TESTS
 
-  To run Cloud V2 tests, execute `tox -e cloud_tests`.
+  To run Cloud V2 tests, execute `tox -e cloud_test_logic`.
```
pandu-k: this makes more sense with the original `tox -e cloud_tests` (as we are running cloud tests, not just the cloud test logic). 
Is this possible, or are these names coupled together?

tests/marqo_test.py
```diff
@@ -180,75 +178,6 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
-class InstanceMappingIndexData(NamedTuple):
```
pandu-k: Nice. I agree moving these out to `cloud_test_logic`

tests/v0_tests/test_bulk_search.py
pandu-k: ahhh sorry. I thought this was the previous function. I needed to expand the diff

tests/v0_tests/test_bulk_search.py
pandu-k: my bad again

tests/v0_tests/test_bulk_search.py
```diff
@@ -667,7 +666,92 @@ def test_query_validation_two_same_urls_success(self):
...
+            "field1": "other things",
+            "multiply": 2,
+            "add": 1,
+            "_id": "123456"
+        }
+
+        self.client.index(self.generic_test_index_name).add_documents([d1], tensor_fields=["doc title"])
+        self.client.index(self.generic_test_index_name_2).add_documents([d2], tensor_fields=["doc title"])
+
+        for search_method in [enums.SearchMethods.TENSOR]:
```
pandu-k: can we do `in [enums.SearchMethods.TENSOR, enums.SearchMethods.LEXICAL]:` while we're here?

tests/marqo_test.py
danyilq: yea it is a typo which happened during refactoring, will change

tests/v0_tests/test_marqo_cloud_instance_mapping.py
pandu-k: 50/50 (doesn't have to be perfect)

Yeah but what if our enum logic breaks?

tests/marqo_test.py
pandu-k: Sorry can you break explain this to me in a simpler way?

tests/marqo_test.py
pandu-k: Can you break up your run on sentence into shorter simpler sentences? It will be easier to follow then.

tests/v0_tests/test_get_settings.py
pandu-k: You are correct

tests/marqo_test.py
danyilq: 1. We execute some test that uses mock_http_traffic decorator and has forbid_extra_calls=True with is_marqo_cloud set to True.
2. If we don’t add any line for catching /indexes request it will not match path and on the check for forbid_extra_calls on the next line it will raise error.

tests/marqo_test.py
danyilq: It's not necessarily an issue limited to tests; it can also occur in other scenarios. For example, when using the with_documents decorator, it initially calls create_test_index and then adds documents to it.
After that, another function is called, which also invokes create_test_index to retrieve the index name it should point to.
If we perform a cleanup operation at this point, any documents created within the with_documents decorator will be deleted.

tests/marqo_test.py
```diff
@@ -248,32 +321,48 @@ def create_test_index(
...
+            cloud_test_index_to_use (Union[CloudTestIndex, None]): The cloud test index to use in cloud environments.
+                If None, an error is raised.
+            open_source_test_index_name (Union[str, None]): The name of the open-source test index to create in
+                open-source environments. If None, no open-source index is created.
+            open_source_index_settings_dict (dict, optional): Additional settings to apply when creating
+                the open-source test index.
+            open_source_index_kwargs (dict, optional): Additional keyword arguments to pass when creating
+                the open-source test index.
+            delete_index_documents_before_test (bool, optional): If True, existing documents in the index will
+                be deleted before preparing it for testing. Default is True. Used only for cloud testing.
```
pandu-k, line 270: Add the explanation of when and why this will be False.

danyilq (commented):
trigger tests

tests/cloud_test_logic/cloud_instance_mappings.py
```diff
@@ -43,16 +43,19 @@ def mock_get_indexes_response(indexes_list: Union[List[GetIndexesIndexResponseOb
     This decorator is used to mock the behavior of the requests.get which is used by
     MarqoCloudInstanceMappings object to retrieve and store specific information about
     cloud indexes: index_name, index status and index endpoint.
-    It allows you to set up mock responses for requests to the "/indexes" URL
-    while allowing real requests to other URLs.
+    It allows you to set up mock responses for requests to the "/indexes" URL.
+    Requests will be handled by side_effect function, which will return mock_get object
+    if url ends with "/indexes", and will call requests.get otherwise.
+    mock_get object can be argument of the decorated function, if to_return_mock is True.
```
pandu-k, line 50: by `mock_get object` do you mean `mocked requests.get response object`?

tests/cloud_test_logic/cloud_instance_mappings.py
```diff
@@ -43,16 +43,19 @@ def mock_get_indexes_response(indexes_list: Union[List[GetIndexesIndexResponseOb
...
         indexes_list (Union[List[GetIndexesIndexResponseObject], None]): A list of
             InstanceMappingIndexData objects to be used for mocking responses
-            to requests to the "/indexes" URL. If None, no mock data will be used.
+            to requests to the "/indexes" URL. If None, mock object will not
+            have return_value set for it.
         to_return_mock (bool): Indicates whether the decorated test function
-            should return the mock object or not. If True, the test function
-            will be called with the mock object as an argument; otherwise, it
+            should return the mock_get object or not. If True, the test function
+            will be called with the mock_get object as an argument; otherwise, it
```
pandu-k, line 60: ditto regarding above comment

tests/marqo_test.py
```diff
@@ -278,6 +275,11 @@ def create_test_index(
         Raises:
             ValueError: If 'cloud_test_index_to_use' is None in cloud environments.
 
+        Notes:
+            - 'delete_index_documents_before_test' is set to True by default in cloud testing to ensure a clean
+              environment. However, in some scenarios, such as when using the with_documents decorator,
+              it may be desirable to set it to False. This allows existing documents to persist when calling
+              'create_test_index' multiple times in the same test.
```
pandu-k, line 282: This is awesome! much cleare. Thanks

tests/v0_tests/test_bulk_search.py
```diff
@@ -68,11 +69,15 @@ def strip_marqo_fields(doc, strip_id=True):
     @mock_get_indexes_response([GetIndexesIndexResponseObject.get_default_index_object()])
     def test_bulk_search_with_context(self):
         """Check that context is passed to HTTP request correctly"""
+        marqo_url_and_version_cache.clear()
+        cache_url = self.client.config.instance_mapping.get_index_base_url(self.generic_test_index_name)
+        marqo_url_and_version_cache[cache_url] = "v0"
```
pandu-k, line 76: add a comment about why you are doing this, within the context of the test (we are simulating that the version is already checked, for example.

tests/v0_tests/test_bulk_search.py
```diff
@@ -114,7 +119,11 @@ def test_bulk_search_with_context(self):
     @mock_get_indexes_response([GetIndexesIndexResponseObject.get_default_index_object()])
     def test_bulk_search_with_scoreModifiers(self):
         """Check that context is passed to HTTP request correctly"""
-        self.client.bulk_search([{
+        marqo_url_and_version_cache.clear()
+        cache_url = self.client.config.instance_mapping.get_index_base_url(self.generic_test_index_name)
+        marqo_url_and_version_cache[cache_url] = "v0"
+        client = marqo.Client(**self.client_settings)
```
pandu-k, line 129: ditto comment above

tests/v0_tests/test_marqo_cloud_instance_mapping.py
```diff
@@ -554,4 +554,10 @@ def test_is_index_usage_allowed_combinations(self):
         assert mapping.is_index_usage_allowed("index3")
         assert mapping.is_index_usage_allowed("index2")
 
+    def test_enum_matches_expected_string(self):
```
pandu-k, line 557: nice. Yay!!

tests/cloud_test_logic/cloud_instance_mappings.py
pandu-k: This is clearer. It can still be improved slightly:

```
if url ends with "/indexes" it returns a mock_get object, otherwise `requests.get` is called normally. 
A mock_get object is a MagicMock object with its return_value set to indexes_list.
```

pandu-k (approved):

PR #168, opened 2023-08-29
deprecate key arguments

danyilq:

* **Other information**:


src/marqo/client.py
```diff
@@ -103,6 +109,11 @@ def create_index(
             storage_node_count:
             replicas_count:
```
pandu-k: Add a comment that indicates the deprecated params are deprecated

src/marqo/index.py
```diff
@@ -108,20 +116,29 @@ def create(config: Config, index_name: str,
             settings_dict: if specified, overwrites all other setting
                 parameters, and is passed directly as the index's
                 index_settings
-            inference_node_type: inference type for the index
-            storage_node_type: storage type for the index
-            inference_node_count: number of inference nodes for the index
-            storage_node_count: number of storage nodes for the index
-            replicas_count: number of replicas for the index
+            inference_node_type (deprecated): inference type for the index
```
pandu-k: For each of these deprecated parameter, indicate which parameter replaces it. For example:

```
inference_node_type (deprecated): inference type for the index. This parameter is replaced by inference_type.
```

src/marqo/utils.py
```diff
@@ -66,3 +71,68 @@ def convert_dict_to_url_params(d: dict) -> str:
...
+    for passed_parameter, default_parameter in passed_and_default:
+        if default_parameter.default is not inspect.Parameter.empty:
+            if arguments_to_ignore and passed_parameter[0] in arguments_to_ignore:
+                continue
+            if passed_parameter[1] != default_parameter.default:
+                return False
+    return True
+
+
+def use_one_of_two_arguments(func, value_name, value, deprecated_value, deprecated_value_name):
```
pandu-k: the order of these params doesn't make 100% sense 

Doesn't the following make more sense? 
```
value_name, value, deprecated_value_name, deprecated_value
```

src/marqo/utils.py
```diff
@@ -66,3 +71,68 @@ def convert_dict_to_url_params(d: dict) -> str:
...
+    """ Checks if both value and deprecated_value are passed to the function and returns only one of them,
+    prioritising non-deprecated one if so. Raises error if both of them are specified"""
+    default_parameters = inspect.signature(func).parameters
+    value_is_specified = value != default_parameters[value_name].default
+    deprecated_value_is_specified = deprecated_value != default_parameters[deprecated_value_name].default
+    if value_is_specified and deprecated_value_is_specified:
+        raise ValueError(f"{value_name} and {deprecated_value_name} can't be specified at the same time "
+                         f"because they refer to the same value, please use only one at a time")
```
pandu-k: We should encourage best usage. If they specify both, tell users to use the new parameter. For example:
```
Parameters `inference_types` and deprecated `inference_node_types` can't be specified at the same time because they refer to the same value. Please only use the `inference_types` parameter. 
```

src/marqo/utils.py
```diff
@@ -66,3 +71,68 @@ def convert_dict_to_url_params(d: dict) -> str:
...
+            if arguments_to_ignore and passed_parameter[0] in arguments_to_ignore:
+                continue
+            if passed_parameter[1] != default_parameter.default:
+                return False
+    return True
+
+
+def use_one_of_two_arguments(func, value_name, value, deprecated_value, deprecated_value_name):
+    """ Checks if both value and deprecated_value are passed to the function and returns only one of them,
+    prioritising non-deprecated one if so. Raises error if both of them are specified"""
```
pandu-k: So the possible failure scenario is this:

```
my_func(new_param = 1, deprecated_param = 2):
      do_something() 

# user's code: 
my_func(new_param = 1, deprecated_param = 3)
# deprecated_param = 3 will be used instead of the expected new_param = 1
```

src/marqo/utils.py
```diff
@@ -66,3 +71,68 @@ def convert_dict_to_url_params(d: dict) -> str:
...
+    default_parameters = inspect.signature(func).parameters
+    value_is_specified = value != default_parameters[value_name].default
+    deprecated_value_is_specified = deprecated_value != default_parameters[deprecated_value_name].default
+    if value_is_specified and deprecated_value_is_specified:
+        raise ValueError(f"{value_name} and {deprecated_value_name} can't be specified at the same time "
+                         f"because they refer to the same value, please use only one at a time")
+    return value if value_is_specified else deprecated_value
```
pandu-k: what if neither are specified?? This looks like it will return deprecated value, which is bad behaviour! This is why we need test cases

src/marqo/models.py
```diff
@@ -24,9 +26,134 @@ class SearchBody(BaseMarqoModel):
...
+    "storage_node_type": "storage_class",
+    "inference_node_count": "number_of_inferences",
+    "storage_node_count": "number_of_shards",
+    "replicas_count": "number_of_replicas",
+}
+
+params_that_can_be_specified_with_settings_dict = ["wait_for_readiness", "settings_dict"]
+
+
+class CreateIndexSettings:
```
pandu-k: you can actually subclass a Pydantic basemodel which provides some boilerplate funtionality. maybe you'd find it helpful? For example: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/models/add_docs_objects.py#L35

src/marqo/models.py
```diff
@@ -24,9 +26,134 @@ class SearchBody(BaseMarqoModel):
...
 
+
 class BulkSearchBody(SearchBody):
     index: str
 
 
 class BulkSearchQuery(BaseMarqoModel):
     queries: List[BulkSearchBody]
+
+
```
pandu-k: Let's move this to a `models/create_index_params.py` file or similar

src/marqo/models/create_index_settings.py
```diff
@@ -0,0 +1,119 @@
...
+    number_of_replicas: int = 0
+
+    def __init__(self, **passed_parameters):
+        passed_parameters_without_none_values = {
+            key: value for key, value in passed_parameters.items() if value is not None
+        }
+        super().__init__(**passed_parameters_without_none_values)
+
+    @validator("*", pre=True)
+    def use_passed_parameters(cls, v, field):
```
pandu-k: what does this validator even do?

src/marqo/models/create_index_settings.py
```diff
@@ -0,0 +1,119 @@
...
+        return v
+
+    @root_validator(pre=True)
+    def handle_deprecated_parameters(cls, values):
+        deprecated_parameters_passed = [
+            deprecated_parameter for deprecated_parameter in index_defaults_deprecated_to_new_params_mapping.keys()
```
pandu-k: you shouldn't need
`deprecated_parameter in index_defaults_deprecated_to_new_params_mapping.keys()`

This should be enough:
`deprecated_parameter in index_defaults_deprecated_to_new_params_mapping`

src/marqo/models/create_index_settings.py
```diff
@@ -0,0 +1,119 @@
...
+                deprecated_parameter
+            )
+        return values
+
+    @root_validator(pre=True)
+    def check_settings_dict_compatibility(cls, values):
+        if values.get("settings_dict") is not None and any(
+                [arg_name for arg_name in values if arg_name not in params_that_can_be_specified_with_settings_dict]
+        ):
+            raise ValueError("settings_dict cannot be specified with other parameters.")
```
pandu-k: add: "... besides wait_for_readiness."

src/marqo/models/create_index_settings.py
```diff
@@ -0,0 +1,119 @@
...
+            if deprecated_parameter in values
+        ]
+        if deprecated_parameters_passed:
+            cls.raise_deprecated_warning(deprecated_parameters_passed)
+        for deprecated_parameter in deprecated_parameters_passed:
+            if index_defaults_deprecated_to_new_params_mapping[deprecated_parameter] in values:
+                raise ValueError(
+                    f"Both {index_defaults_deprecated_to_new_params_mapping[deprecated_parameter]} and "
+                    f"it's deprecated reference {deprecated_parameter} "
+                    f"are specified. Please consider using only "
```
pandu-k: Let's be more authoritative. If the user specifies both we want them to choose the non-deprecated version: `Please consider using only` -> `Please only use ...`

danyilq (commented):
trigger tests

src/marqo/client.py
```diff
@@ -91,6 +91,11 @@ def create_index(
             number_of_replicas=None
 ) -> Dict[str, Any]:
         """Create the index. Please refer to the marqo cloud to see options for inference and storage node types.
+        Calls Index.create() with the same parameters.
+        All parameters are optional, and will be set to their default values if not specified.
+        Default values can be found in models/create_index_settings.py CreateIndexSettings class.
+
+
```
pandu-k, line 98: nit too many newlines

src/marqo/models/create_index_settings.py
```diff
@@ -24,13 +24,16 @@ class CreateIndexSettings(BaseModel):
...
-            wait_for_readiness (bool): Marqo Cloud specific, whether to wait until the operation is completed or to proceed
-            without waiting for status; won't do anything if config.is_marqo_cloud=False.
+            inference_node_type (str, deprecated): Inference type for the index. Deprecated, use inference_type instead.
+            storage_node_type (str, deprecated): Storage type for the index. Deprecated, use storage_class instead.
+            inference_node_count (int, deprecated): Number of inference nodes for the index. Deprecated,
+            use number_of_inferences instead.
+            storage_node_count (int, deprecated): Number of storage nodes for the index. Deprecated,
+            use number_of_shards instead.
+            replicas_count (int, deprecated): Number of replicas for the index. Deprecated,
+            use number_of_replicas instead.
```
pandu-k, line 34: nit, indent the overflowing line:
```
            inference_node_count (int, deprecated): Number of inference nodes for the index. Deprecated,
                use number_of_inferences instead.
```

src/marqo/models/create_index_settings.py
```diff
@@ -76,37 +79,53 @@ def __init__(self, **passed_parameters):
...
-        return v
-
     @root_validator(pre=True)
     def handle_deprecated_parameters(cls, values):
+        """ Handles deprecated parameters and raises deprecation warnings.
+        Ensures that deprecated parameters are not specified along with their new counterparts.
+        If both deprecated and new parameters are specified, raises a ValueError.
+        If only deprecated parameters are specified, sets new parameters to the values of deprecated parameters.
+        DEPRECATED_TO_NEW_PARAMS_MAPPING is a dictionary that maps deprecated parameters to
+        their new counterparts.
```
pandu-k, line 89: nit: this is good, but it is better to describe the intention and side effects of the function than repeating the logic. This is subtle distinction

src/marqo/models/create_index_settings.py
```diff
@@ -76,37 +79,53 @@ def __init__(self, **passed_parameters):
...
+
+        Notes:
+            if you want to add a new deprecated parameter, add it to DEPRECATED_TO_NEW_PARAMS_MAPPING.
```
pandu-k, line 92: helpful note!

src/marqo/index.py
```diff
@@ -165,19 +170,29 @@ def create(config: Config, index_name: str,
...
+            else:
+                cl_img_preprocessing.pop('patch_method', None)
             if not config.is_marqo_cloud:
                 return req.post(f"indexes/{index_name}", body=cl_settings)
-            cl_settings['inference_type'] = create_index_settings.inference_type
-            cl_settings['storage_class'] = create_index_settings.storage_class
+            if create_index_settings.inference_type is not None:
+                cl_settings['inference_type'] = create_index_settings.inference_type
+            else:
+                cl_settings.pop('inference_node_type', None)
```
pandu-k: why do we pop `inference_node_type?`

src/marqo/models/create_index_settings.py
```diff
@@ -76,37 +79,53 @@ def __init__(self, **passed_parameters):
...
+        Notes:
+            if you want to add a new parameter that can be specified along with settings_dict, add it to
+            SETTINGS_DICT_COMPATIBLE_PARAMS.
+        """
         if values.get("settings_dict") is not None and any(
-                [arg_name for arg_name in values if arg_name not in params_that_can_be_specified_with_settings_dict]
+                [arg_name for arg_name in values if arg_name not in SETTINGS_DICT_COMPATIBLE_PARAMS]
         ):
-            raise ValueError("settings_dict cannot be specified with other parameters.")
+            raise ValueError(f"settings_dict cannot be specified with other parameters. "
```
pandu-k: wording. Change to:
`other index creation parameters, besides `

src/marqo/index.py
```diff
@@ -165,19 +170,29 @@ def create(config: Config, index_name: str,
...
             if image_preprocessing_method is not None:
                 cl_img_preprocessing['patch_method'] = create_index_settings.image_preprocessing_method
+            else:
+                cl_img_preprocessing.pop('patch_method', None)
             if not config.is_marqo_cloud:
                 return req.post(f"indexes/{index_name}", body=cl_settings)
-            cl_settings['inference_type'] = create_index_settings.inference_type
-            cl_settings['storage_class'] = create_index_settings.storage_class
+            if create_index_settings.inference_type is not None:
+                cl_settings['inference_type'] = create_index_settings.inference_type
```
pandu-k, line 189: do we need logic to ensure we set this to either `inference_type` XOR `inference_node_type`? same goes for storage and counts

pandu-k: We may need to add new methods to CreateIndexSettings that gets the correct attribute

danyilq: during index config creation if deprecated param is passed and None is passed for it's new param we simply state that new parameter equals passed deprecated value and we never set the deprecated parameter

tests/v0_tests/test_create_index_settings.py
```diff
@@ -0,0 +1,104 @@
...
+            storage_node_type="marqo.basic",
+            replicas_count=1,
+            inference_node_count=2,
+            storage_node_count=2,
+        )
+        assert settings.inference_type == "marqo.CPU"
+        assert settings.storage_class == "marqo.basic"
+        assert settings.number_of_inferences == 2
+        assert settings.number_of_shards == 2
+        assert settings.number_of_replicas == 1
```
pandu-k, line 73: nit: we should assert that the deprecated attributes _don't_ get set.

pandu-k (approved):

PR #189, opened 2023-12-20
Add Cloud Support for Marqo V2

wanliAlex:
Add cloud support for Marqo V2:
1. Add `v2` prefix for all `indexes/` APIs for control panel.
2. Add pydantic models for cloud index requests.
3. Fix the camel case in cloud response


src/marqo/_httprequests.py
```diff
@@ -37,7 +38,14 @@ def _construct_path(self, path: str, index_name="") -> str:
         """Augment the URL request path based if telemetry is required."""
         base_url = self.config.instance_mapping.get_index_base_url(index_name=index_name) if index_name \
             else self.config.instance_mapping.get_control_base_url()
+
+        # Add v2 prefix if the request is sent to controller index API
+        if base_url.lower().startswith(os.environ.get("MARQO_CLOUD_URL", "https://api.marqo.ai")) and \
+                path.startswith("indexes"):
+            path = f"v2/{path}"
```
pandu-k: nit: not a huge fan of mutating this var

would prefer something like 
```
if base_url...
    # control plane endpoints are now versioned
    url = f"{base}/v2/{path}"
else:
    # data plane endpoints are not 
    url = f"{base}/{path}"
```

src/marqo/index.py
```diff
@@ -119,68 +127,85 @@ def create(config: Config,
...
+                    imagePreprocessing=image_preprocessing,
+                    vectorNumericType=vector_numeric_type,
+                    annParameters=ann_parameters,
+                    numberOfInferences=number_of_inferences,
+                    inferenceType=inference_type,
+                    numberOfShards=number_of_shards,
+                    numberOfReplicas=number_of_replicas,
+                    storageClass=storage_class,
+                )
+            # client instantiated with endpoint url
```
pandu-k: note: this should not work. it could be OK to still send this request to cloud, as the error will come back from the cloud, too. Please test it and check what the error is (the error should make sense, for example: 409 Conflict)

src/marqo/models/create_index_settings.py
```diff
@@ -11,6 +11,8 @@ class IndexSettings(StrictBaseModel):
         type: The type of the index. Can be structured or unstructured.
         allFields: A list of all fields in the index.
         tensorFields: A list of all tensor fields in the index.
+        treatUrlsAndPointersAsImages: Whether to treat urls and pointers as images.
```
pandu-k, line 16: nit: it could be worth stating in the doc string that this is an unstructured-only param

src/marqo/models/create_index_settings.py
```diff
@@ -26,8 +28,10 @@ class IndexSettings(StrictBaseModel):
 
     type: Optional[marqo_index.IndexType] = Field(None, alias="type")
     allFields: Optional[List[marqo_index.FieldRequest]] = Field(None, alias="all_fields")
+    settingsDict: Optional[Dict] = Field(None, alias="settings_dict")
     tensorFields: Optional[List[str]] = Field(None, alias="tensor_fields")
     treatUrlsAndPointersAsImages: Optional[bool] = Field(None, alias="treat_urls_and_pointers_as_images")
+    shortStringLengthThreshold: Optional[int] = Field(None, alias="short_string_length_threshold")
```
pandu-k: what do these Aliases do?

src/marqo/models/create_index_settings.py
```diff
@@ -36,26 +40,22 @@ class IndexSettings(StrictBaseModel):
...
     def check_settings_dict_compatibility(cls, values):
-        if values.get("settings_dict") is not None and \
-                values.get("index_settings").dict(exclude_none=True) != {}:
-            raise ValueError("settings_dict and index_settings cannot be used together")
-        return values
+        """ Ensures that settingsDict is not specified along with other parameters.
+
+        Notes:
+            if you want to add a new parameter that can be specified along with settings_dict, add it to
+            SETTINGS_DICT_COMPATIBLE_PARAMS.
```
pandu-k: does adding WAIT_FOR_READINESS to SETTINGS_DICT_COMPATIBLE_PARAMS solve the problem you mentioned?

src/marqo/models/marqo_cloud.py
```diff
@@ -0,0 +1,31 @@
...
+    numberOfShards: Optional[int] = Field(None, alias="number_of_shards")
+    numberOfInferences: Optional[int] = Field(None, alias="number_of_inferences")
+
+
+class ListIndexesResponse(CloudIndexSettings):
+    Created: Optional[str] = None
+    indexName: Optional[str] = None
+    errorMsg: Optional[str] = None
+    marqoEndpoint: Optional[str] = None
+    index_defaults: Optional[dict] = None
```
pandu-k, line 25: any update from Ali or Vivien on this?

src/marqo/_httprequests.py
wanliAlex: Done. Thanks!

src/marqo/models/create_index_settings.py
wanliAlex: done

src/marqo/models/create_index_settings.py
wanliAlex: Agreed, we do not need these alias anymore

src/marqo/index.py
wanliAlex: Do you mean we don't need to branch at all? We can just send the full body and rely on the cloud to return errors?

src/marqo/models/marqo_cloud.py
wanliAlex: This should be deleted. Confirmed with Vivien

src/marqo/models/create_index_settings.py
wanliAlex: Will discuss this with you in person. I think the behaviour in v1 is not expected.

wanliAlex: The parameter `wait_for_readiness` should not be included in the `settings_dict` anymore as `settings_dict` is a dictionary containing only index-related parameters, e.g., users copy paste from their curl request.

src/marqo/_httprequests.py
pandu-k: @wanliAlex did you test this?

wanliAlex: Yes, shallow tests

pandu-k (approved):

src/marqo/_httprequests.py
```diff
@@ -37,7 +38,16 @@ def _construct_path(self, path: str, index_name="") -> str:
         """Augment the URL request path based if telemetry is required."""
         base_url = self.config.instance_mapping.get_index_base_url(index_name=index_name) if index_name \
             else self.config.instance_mapping.get_control_base_url()
-        url = f"{base_url}/{path}"
+
+        # Add v2 prefix if the request is sent to controller index API
+        if base_url.lower().startswith(os.environ.get("MARQO_CLOUD_URL", "https://api.marqo.ai")) and \
```
farshidz: This isn't the righrt place for this. `MarqoCloudInstanceMappings` is where this kind of logic should go

src/marqo/models/marqo_cloud.py
```diff
@@ -0,0 +1,32 @@
...
+class CloudIndexSettings(IndexSettings):
+    inferenceType: Optional[str] = Field(None, alias="inference_type")
+    storageClass: Optional[str] = Field(None, alias="storage_class")
+    numberOfReplicas: Optional[int] = Field(None, alias="number_of_replicas")
+    numberOfShards: Optional[int] = Field(None, alias="number_of_shards")
+    numberOfInferences: Optional[int] = Field(None, alias="number_of_inferences")
+
+
+class ListIndexesResponse(CloudIndexSettings):
+    Created: Optional[str] = None
```
farshidz: Why is this one PascalCase?

src/marqo/models/marqo_cloud.py
```diff
@@ -0,0 +1,32 @@
...
+    indexName: Optional[str] = None
+    errorMsg: Optional[str] = None
+    marqoEndpoint: Optional[str] = None
+    # TODO Remove this attribute after cloud is updated
+    index_defaults: Optional[dict] = None
+    marqoVersion: Optional[str] = None
+    indexStatus: Optional[str] = None
+
+
+class IndexResponseEnum(str, Enum):
```
farshidz: I don't think this is correct usage of enum. Why don't you create a model for status endpoint response so you can deserialize the response properly?

src/marqo/client.py
```diff
@@ -108,6 +110,16 @@ def create_index(
             image_preprocessing: image preprocessing settings
             vector_numeric_type: vector numeric type
             ann_parameters: approximate nearest neighbors parameters
+            # Below are cloud specific parameters
```
farshidz: I don't think this line is valid syntax for Args. It's important to use the right syntax so IDEs can parse this documentation

src/marqo/index.py
```diff
@@ -119,68 +124,67 @@ def create(config: Config,
             image_preprocessing: image preprocessing settings
             vector_numeric_type: vector numeric type
             ann_parameters: approximate nearest neighbors parameters
+            # Below are cloud specific parameters
```
farshidz: Comment in args in unconventional and not ideal

src/marqo/index.py
```diff
@@ -119,68 +124,67 @@ def create(config: Config,
...
-        #     cl_settings['number_of_replicas'] = create_index_settings.number_of_replicas
-        #     cl_settings['number_of_shards'] = create_index_settings.number_of_shards
-        #     response = req.post(f"indexes/{index_name}", body=cl_settings)
-        #     if create_index_settings.wait_for_readiness:
-        #         cloud_wait_for_index_status(req, index_name, IndexStatus.READY)
-        #     return response
-        return req.post(f"indexes/{index_name}", body=create_index_settings.request_body)
+
+        # py-marqo against local Marqo
+        if not config.api_key:
```
farshidz: This will eval to true if `api_key` is an empty string. I think you should check for `is None` instead

src/marqo/index.py
```diff
@@ -119,68 +124,67 @@ def create(config: Config,
...
+                textPreprocessing=text_preprocessing,
+                imagePreprocessing=image_preprocessing,
+                vectorNumericType=vector_numeric_type,
+                annParameters=ann_parameters
+            )
+
+            return req.post(f"indexes/{index_name}", body=local_create_index_settings.generate_request_body())
+
+        # py-marqo against Marqo Cloud
+        elif config.api_key:
```
farshidz: No `else`?

farshidz (changes requested):

src/marqo/models/marqo_cloud.py
wanliAlex: This is a problem in the cloud and they decided not to fix it. I can make it camelCase and alias it.

src/marqo/client.py
wanliAlex: Move the comment to `Note`

src/marqo/index.py
wanliAlex: done

src/marqo/index.py
wanliAlex: fixed

src/marqo/models/marqo_cloud.py
wanliAlex: good idea

wanliAlex: done

src/marqo/_httprequests.py
wanliAlex: Thanks for catching this. Fixed

wanliAlex (commented):
Update branch based on Farshid's comments

src/marqo/_httprequests.py
```diff
@@ -1,5 +1,6 @@
 import copy
 import json
+import os
```
farshidz: This import isn't used. You can get PyCharm to organize and clean up imports before committing

farshidz (approved):

farshidz (approved):

PR #192, opened 2023-12-29
Fix github workflow on opensource/cloud tests

wanliAlex:


wanliAlex:
image


This is the automatically triggered workflow. The py-marqo test is running against the expected Marqo docker image.

PR #197, opened 2024-01-03
Li/fix cloudtests pipeline

wanliAlex:
Fix cloud integration pipeline, opensource pipeline, and add create_indexes tests


tests/cloud_test_logic/cloud_test_index.py
```diff
@@ -6,65 +6,112 @@ class CloudTestIndex(str, Enum):
...
+    # },
+    #Structured indexes
+    CloudTestIndex.structured_basic_index: {
+        "type": "structured",
+        "inferenceType": "marqo.CPU.large",
+        "model": "sentence-transformers/all-MiniLM-L6-v2",
+        "allFields": [{"name": "text_field_1", "type": "text"}],
+        "tensorFields": ["text_field_1"],
+        "storageClass": "marqo.basic"
     },
```
pandu-k: so we only have 2 indexes. I am guessing this is due to resource limits right? Are you sure this hasn't been resolved? 

At the very least can one one index be an image index please?

pandu-k: Also, can each index use different infra ? So like one should use GPUs and one should use a balanced storage nodes. 

Also, one should have replicas and one should not

pandu-k: and at least 1 index with more than 1 inference node

tests/cloud_test_logic/populate_indices_for_cloud_tests.py
```diff
@@ -4,6 +4,8 @@
 import marqo
 from cloud_test_index import index_name_to_settings_mappings
 
+INDEX_NAME_SEPARATOR = "_"
```
pandu-k, line 7: nit: should we have this as an env var - so it can be used in all the logic. this is just a nice to have though, and it shouldn't block the merge of this PR

tests/conftest.py
pandu-k, line 1: nit - a quick docstring at the top of this file would be helpful

tests/v0_tests/test_create_index_settings.py
pandu-k, line 1: qq: why was this deleted?

tests/marqo_test.py
pandu-k, line 1: Do we need to update the docstring at the top of this file - especially around the `--all` option?
(also - note for future work - this docstring should be its own test-suite specific README file)

tests/marqo_test.py
```diff
@@ -199,40 +202,22 @@ def setUpClass(cls) -> None:
...
-                    if index.index_name.endswith(cls.index_suffix):
-                        cls.cleanup_documents_from_index(cls, index.index_name)
-
+        if cls.client.config.is_marqo_cloud:
+            for index in cls.client.get_indexes()['results']:
+                if index.index_name.endswith(cls.index_suffix):
+                    cls.cleanup_documents_from_index(cls, index.index_name)
+        else:
+            if cls.open_source_indexes_list:
+                cls.delete_open_source_indexes(cls.open_source_indexes_list)
```
pandu-k, line 218: thanks for refactoring this - it makes much more sense like this XD

tests/marqo_test.py
```diff
@@ -322,27 +306,44 @@ def prepare_cloud_index_for_test(self, index_name: str, delete_index_documents_b
...
+            r.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            raise MarqoWebError(e)
+
+    @classmethod
+    def clear_open_source_indexes(cls, index_names: List[str]):
+        if cls.client.config.is_marqo_cloud:
+            raise MarqoError("clear_open_source_indexes is not supported in cloud environments")
+        for index_name in index_names:
+            r = requests.delete(f"{cls.authorized_url}/indexes/{index_name}/documents/delete-all")
```
pandu-k, line 353: the delete-all endpoint is actually a `POST` endpoint:
image

pandu-k: Are we sure that this method has been running properly if we haven't detected this bug?

FYI - relevant Marqo API line: https://github.com/marqo-ai/marqo/blob/mainline-poc/src/marqo/tensor_search/api.py#L215

tests/marqo_test.py
```diff
@@ -199,40 +202,22 @@ def setUpClass(cls) -> None:
...
-
+        if cls.client.config.is_marqo_cloud:
+            for index in cls.client.get_indexes()['results']:
+                if index.index_name.endswith(cls.index_suffix):
+                    cls.cleanup_documents_from_index(cls, index.index_name)
+        else:
+            if cls.open_source_indexes_list:
+                cls.delete_open_source_indexes(cls.open_source_indexes_list)
 
     def setUp(self) -> None:
```
pandu-k, line 220: it is worth adding a docstring here describing why we only clear open source indexes in setUp

tests/marqo_test.py
```diff
@@ -199,40 +202,22 @@ def setUpClass(cls) -> None:
...
-            for index in self.client.get_indexes()['results']:
-                if index.index_name.startswith(self.generic_test_index_name):
-                    try:
-                        index.delete()
-                    except marqo.errors.MarqoApiError as e:
-                        logging.debug(f'received error `{e}` from index deletion request.')
+        if not self.client.config.is_marqo_cloud and self.open_source_indexes_list:
+            self.clear_open_source_indexes(self.open_source_indexes_list)
 
     def tearDown(self) -> None:
```
pandu-k, line 226: it is worth adding a docstring here describing why we only clear open source indexes in setUp

pandu-k, line 226: Also, when do we clear the cloud indexes?

pandu-k: maybe in this function: `add_documents_and_mark_for_cleanup_patch()`...

tests/marqo_test.py
```diff
@@ -243,29 +228,30 @@ def warm_request(self, func, *args, **kwargs):
         for i in range(5):
             func(*args, **kwargs)
 
-    def create_test_index(
+    def get_test_index_name(
             self,
-            cloud_test_index_to_use: Union[CloudTestIndex, None], open_source_test_index_name: Union[str, None],
-            open_source_index_settings_dict: dict = None, open_source_index_kwargs: dict = None,
+            cloud_test_index_to_use: Union[CloudTestIndex, None],
+            open_source_test_index_name: Union[str, None],
```
pandu-k, line 243: what is the impact of removing the open source settings dict? are we no longer setting up new open source indexes for every class?

tests/marqo_test.py
```diff
@@ -243,29 +228,30 @@ def warm_request(self, func, *args, **kwargs):
...
         In the case of cloud testing, it provides the name of the cloud index to be used.
         Additionally, it applies a unique run identifier to the index name and performs
         cleanup operations on documents associated with the index if 'delete_index_documents_before_test' is True.
 
+        In the case of open-source testing, it will return the open-source index name to be used. The documents
+        in the index are cleaned by 'setUp' function of the test case.
+
+        Note: Index creation for both cloud and open-source environments is NOT included in this function.
+        In the case of cloud testing, the index created at the start of py-marqo test.
+        In the case of open-source testing, the index is created in the 'setUpClass' method of the test case.
```
pandu-k, line 259: this is helpful. thanks!

tests/marqo_test.py
pandu-k, line 1: We definitely need to update the `## NOTES ABOUT THE TEST SUITE BEHAVIOUR ##` part of the docstring to make it reflect the new way the test suite works. Especially regarding, for cloud and open source:
a. When indexes are created and deleted (inc. relevant functions)
b. When documents are are cleared from an index (inc. relevant functions)

tests/marqo_test.py
```diff
@@ -243,29 +228,30 @@ def warm_request(self, func, *args, **kwargs):
...
-            cloud_test_index_to_use: Union[CloudTestIndex, None], open_source_test_index_name: Union[str, None],
-            open_source_index_settings_dict: dict = None, open_source_index_kwargs: dict = None,
+            cloud_test_index_to_use: Union[CloudTestIndex, None],
+            open_source_test_index_name: Union[str, None],
             delete_index_documents_before_test: bool = True
     ):
-        """Determines whether the test is executed in a cloud environment or within an open-source setup.
-        If it's running in an open-source environment,
-        this function creates the corresponding index and returns its name.
+        """Get the test index name for unittest test cases, based on whether the test is running in a cloud environment.
```
pandu-k: This is a good improvement to the docstring's clarity. We can even be clearer though:

`Gets the name of an index that is used in a given unittest test case, based on whether the test is running in a cloud environment. The names of indexes on the cloud and in open source are slightly different.`

tests/cloud_test_logic/cloud_test_index.py
pandu-k: And at least one index should have a dash (`-`) in its name and one should have an underscore (`_`) - just to verify this all works.

pandu-k: also. there should be a test where two indexes are the same, but one uses a hyphen and the other uses a dash. For example: 
`my-index`, `my_index`. This is because the index name part of the index's URL will have substitute dashes for underscores (and a random string for uniqueness) - but we should include it as a test to ensure that it works

tests/v2_tests/test__httprequests.py
```diff
@@ -0,0 +1,111 @@
...
+
+        with self.assertRaises(MarqoWebError):
+            http_requests.get('/', index_name="test_index")
+
+        mock_index_http_error_handler.assert_called_once()
+        mock_index_http_error_handler.assert_called_with("test_index")
+
+
+@pytest.mark.fixed
+class TestConstructCloudPath(unittest.TestCase):
```
pandu-k, line 61: nice test case - thanks for adding!

tests/v2_tests/test__httprequests.py
```diff
@@ -0,0 +1,111 @@
...
+            "indexes/", "indexes/my_index_name/documents", "indexes/my_index_name/search",
+            "indexes/my_index_name/stats", "indexes/my_index_name/settings",
+            "indexes/my_index_name/health"
+        ]
+        for path in test_cases:
+            with self.subTest(f"self.cloud_url={self.cloud_url}, path={path}"):
+                result = self.construct_path_helper(self.cloud_url, path)
+                self.assertEqual(f"{self.cloud_url}/api/v2/{path}", result)
+
+    def test_path_not_start_with_indexes(self):
```
pandu-k, line 88: hmmm are sure about this one? because all the dataplane endpoints (like `/search`) should be sent to the specific dataplane URL (like `https://my-indx-abcdef-hijklm.marqo.ai/indexes/my-indexes/search`).

Maybe this isn't doesn't matter too much if it demonstrates the correct logic. But still, a decent amount of our test cases (besides edge cases) should reflect reality

tests/v2_tests/test_create_index.py
pandu-k, line 1: nice - this looks like a significant step up from the previous test

tests/v2_tests/test_create_index_settings.py
pandu-k, line 1: why are these commented out? can't we mark these as unfixed?

tests/v2_tests/test_index_features.py
pandu-k, line 1: 1. I think the name of this file could be more descriptive and more accurate. For example `test_index_manipulation_features` 
2. Most of these should be ignored during cloud tests

pandu-k (approved):

tests/conftest.py
wanliAlex: Added

tests/marqo_test.py
wanliAlex: This is just a temporary argument to run all the tests. It should be remove after all the tests are fixed and there will be not `fixed` mark

tests/v0_tests/test_create_index_settings.py
wanliAlex: I think it's just GitHub issue. I am changing the directory name from v0_tests to v2_tests

tests/v2_tests/test_create_index_settings.py
wanliAlex: done

tests/v2_tests/test__httprequests.py
wanliAlex: Yes, I am testing the logic here.

I have added another test to ensure v2 is not added to dataplane endpoints.

tests/v2_tests/test_index_features.py
wanliAlex: That's a good name.

I haven't fixed the tests here. I only move these index-related tests here. Marks can be added when these tests are fixed.

wanliAlex (commented):
Update PR based on pandu's comments

tests/cloud_test_logic/cloud_test_index.py
wanliAlex: All the comments are addressed except those related to `dash`

tests/marqo_test.py
wanliAlex: done

tests/marqo_test.py
wanliAlex: In function `prepare_cloud_index_for_test`

tests/marqo_test.py
wanliAlex: done

tests/marqo_test.py
wanliAlex: We are not using this function to create open-source index anymore.

tests/marqo_test.py
wanliAlex: We are not using this API here. We call a test-only API that clears all the documents in a given index. No `ids` is required.

tests/marqo_test.py
wanliAlex: added

tests/cloud_test_logic/cloud_test_index.py
```diff
@@ -5,113 +5,117 @@ class CloudTestIndex(str, Enum):
...
+
+    3) custom_model_index_with_text_preprocessing: a text-only index with custom model and text preprocessing.
+     This should be used to test documents with only text fields that requires text preprocessing.
+
+    For examples,
+        1) You want to test text fields without text preprocessing-> use basic_index
+        2) You want to test image fields without image preprocessing-> use basic_index
+        3) You want to test image fields with image preprocessing -> use image_index_with_image_preprocessing
+        4) You want to test text fields with text preprocessing -> use custom_model_index_with_text_preprocessing
+    """
```
pandu-k, line 31: helpful docstring.

tests/cloud_test_logic/cloud_test_index.py
```diff
@@ -5,113 +5,117 @@ class CloudTestIndex(str, Enum):
...
+        "numberOfInferences": 2,
+    },
+    CloudTestIndex.unstructured_image_index_with_image_preprocessing: {
+        "type": "unstructured",
+        "treatUrlsAndPointersAsImages": True,
+        "model": "open_clip/ViT-B-32/laion400m_e32",
+
         "inferenceType": "marqo.CPU.large",
+        "storageClass": "marqo.balanced",
+        "numberOfReplicas": 2,
```
pandu-k: you can only have 1 replica

pandu-k: the cloud should validate against it

tests/cloud_test_logic/cloud_test_index.py
pandu-k: Also, we can get even more variations if we distribute test cases across structured/unstructured. For example:
custom-text-model: unstructured
custom-image-model: structured

Of course this means we don't cover every permutation, but it isn't practical to cover every possible combination during integ tests (and this is the realm of unit tests). We should instead aim to cover a broad set of use cases.

tests/cloud_test_logic/cloud_test_index.py
pandu-k, line 1: to increase coverage: ideally every index uses a slightly different model, ideally from different architectures (we can ignore rarely used ones though)

pandu-k: we should look to maximise the entropy of the index settings to broaden our testing net

tests/cloud_test_logic/cloud_test_index.py
pandu-k: ideally, every index also has slightly different infrastructure configurations too

tests/cloud_test_logic/cloud_test_index.py
```diff
@@ -5,113 +5,117 @@ class CloudTestIndex(str, Enum):
...
-    #     "type": "unstructured",
-    #     "treatUrlsAndPointersAsImages": True,
-    #     "model": "open_clip/ViT-B-16/laion400m_e31",
-    #     "imagePreprocessing": {
-    #         "patchMethod": "simple"
-    #     },
-    #     "inferenceType": "marqo.CPU.large",
-    #     "storageClass": "marqo.basic"
-    # },
+
```
pandu-k: we need more entropy in the selected fields for structured indexes:
e.g. (add even more entropy if you can think of it!):
- bools
- floats
- ints
- fields used for filtering
- fields used for score modification
- field _not used for filtering_
- fields used for lexical
- text fields _not_ used for lexical
-

tests/marqo_test.py
```diff
@@ -52,19 +52,26 @@
...
 
-When `cluster.is_marqo_cloud` is set to True, the tests will have different setUp and tearDown procedures:
-- Indices will not be deleted after each test.
-- However, documents will be deleted on every call for index.
+The index creation and deletion is handled differently for cloud and open-source tests.
 
+
+1) Cloud Environment:
+When `cluster.is_marqo_cloud` is set to True:
+- All the indexes will be created at the start of the test suite, in file
```
pandu-k: nit: I prefer `test run` over `test suite` as "suite" is more about the code and "run" evokes a program in flight

tests/marqo_test.py
```diff
@@ -52,19 +52,26 @@
...
+marqo.tests.populate_indices_for_cloud_tests.py
+- Indices will not be deleted until the end of the workflow
+- However, documents will be deleted on every call for index, in function prepare_cloud_index_for_test.
```
pandu-k: what does "every call for index" mean - this doesn't make much sense to me

tests/marqo_test.py
```diff
@@ -52,19 +52,26 @@
...
 
-We will not actually create and delete real cloud indexes
-during this test suite, because this slows down py-marqo <> Marqo cloud tests.
-We should still test these methods with mocking. End-to-end
-creation and deletion tests will be done elsewhere.
+2) Open-Source Environment:
+When `cluster.is_marqo_cloud` is set to False:
+- All the indexes will be created in the setUpClass method of the test suite. We should only use the
+created indexes in a test class.
+- Indices will not be deleted after each unittest, but will be cleared in the tearDownClass method of the test suite.
```
pandu-k: `after each unittest,` -> `after each test method,`. "unittest` is the name of the package

tests/marqo_test.py
```diff
@@ -52,19 +52,26 @@
...
-add_documents_and_mark_for_cleanup_patch instead, which acts as as a decorator for the original method.
+add_documents_and_mark_for_cleanup_patch instead, which acts as a decorator for the original method.
 This method stores a list of added documents for future cleanup of documents, and then performs the real request.
 
-We will not actually create and delete real cloud indexes
-during this test suite, because this slows down py-marqo <> Marqo cloud tests.
-We should still test these methods with mocking. End-to-end
-creation and deletion tests will be done elsewhere.
+2) Open-Source Environment:
+When `cluster.is_marqo_cloud` is set to False:
```
pandu-k, line 70: nice - great to add this details so that devs understand how the test suite determines whether it will be cloud test

tests/marqo_test.py
```diff
@@ -52,19 +52,26 @@
...
+- All the indexes will be created in the setUpClass method of the test suite. We should only use the
+created indexes in a test class.
+- Indices will not be deleted after each unittest, but will be cleared in the tearDownClass method of the test suite.
+- Documents will be deleted in the setUp method of each test class, before running any unittest.
```
pandu-k: this makes it sound like you are deleted docs before each test class runs. This is clearer: `Documents will be deleted before every test method runs, via the test class' setUp() method.` (notice `every test method` is emphasised by being earlier on in the sentence

tests/marqo_test.py
pandu-k: nit - a docstring is better than just line comments, because there are some programs that parse docstrings and use it for documentation

.github/workflows/cloud-integration-tests.yml
```diff
@@ -17,9 +17,11 @@ on:
...
         required: true
         default: 'run_integration_tests'
-  pull_request_review:
-    types:
-      - submitted
+  pull_request:
+    branches:
+      - mainline
+      - 'releases/*'
+      - mainline-poc  # TODO remove this once we have a mainline branch for V2
```
pandu-k: do you think this is a bit aggressive? I am pretty sure every time you push a new commit to your PR branch it reruns the test suite

tests/v2_tests/test_create_index.py
```diff
@@ -6,10 +6,9 @@
 from tests.marqo_test import MarqoTestCase
 
 
-@mark.fixed
 @mark.ignore_during_cloud_tests
 class TestCreateIndex(MarqoTestCase):
-    index_name = "test_create_index" + str(uuid.uuid4()).replace('-', '')
+    index_name = "test-create_index" + str(uuid.uuid4()).replace('-', '')
```
pandu-k, line 11: should we move index names with dashes to a future PR so that we can get this PR's workflows working properly?

tests/marqo_test.py
wanliAlex: Updated

pandu-k (changes requested):
sorry - accidentally approved before

tests/cloud_test_logic/cloud_test_index.py
wanliAlex: done

tests/cloud_test_logic/cloud_test_index.py
wanliAlex: I have added some fields for structured indexes, but you can add more when adding unittests

tests/marqo_test.py
wanliAlex: updated

tests/marqo_test.py
wanliAlex: done

tests/v2_tests/test_create_index.py
wanliAlex: I removed the mark `fixed` for this test so it will not run during the pipeline. We can add back the mark once the Marqo PR is merged

pandu-k (approved):

pandu-k:
running cloud tests: https://github.com/marqo-ai/py-marqo/actions/runs/7431501634

PR #198, opened 2024-01-03
pin requirements version

wanliAlex:
pin requirments for py-marqo to fix SSL certificate issue

The versions are from `pip freeze`


pandu-k (approved):

pandu-k (changes requested):
.

wanliAlex (commented):
We decided to use range for `urllib3`

pandu-k (approved):

PR #211, opened 2024-01-14
Fix all tests and delete unneded ones

danyilq:

* **Other information**:


danyilq (commented):
trigger cloud tests

tests/cloud_test_logic/cloud_test_index.py
wanliAlex, line 1: Do we reduce the test index diversity to 3 indexes?  It was previously 2 indexes due to the lack of resources in the staging account. We should test more indexes settings in this test.

tests/marqo_test.py
```diff
@@ -213,6 +216,58 @@ def setUpClass(cls) -> None:
...
+                        "type": "structured",
+                        "allFields": [{"name": "text_field_1", "type": "text", "features": ["lexical_search"]},
+                                      {"name": "text_field_2", "type": "text", "features": ["lexical_search"]},
+                                      {"name": "text_field_3", "type": "text", "features": ["lexical_search"]},
+                                      {"name": "image_field_1", "type": "image_pointer"},
+                                      ],
+                        "tensorFields": ["text_field_1", "text_field_2", "text_field_3", "image_field_1"],
+                        "model": "ViT-B/32",
+                    }
+                    # {
```
wanliAlex, line 255: Are we not testing any image preprocessing in the py-marqo test?

tests/marqo_test.py
```diff
@@ -213,6 +216,58 @@ def setUpClass(cls) -> None:
...
+                    #     "treatUrlsAndPointersAsImages": True,
+                    # },
+                ])
+            except Exception as e:
+                print("Error creating indexes: ", e)
+
+
     @classmethod
     def tearDownClass(cls) -> None:
         """Delete commonly used test indexes after all tests are run
```
wanliAlex, line 273: If we use the same index settings for all the test classes, we don't have to create/delete indexes in setUpClass and tearDownClass even in open source.

Confirm with @pandu-k if this kind of setting is acceptable. On the one hand, we have fewer tests for index creation. On the other hand, we can leave that responsibility to API tests.

tests/v2_tests/test_create_index.py
```diff
@@ -6,9 +6,10 @@
 from tests.marqo_test import MarqoTestCase
 
 
+@mark.fixed
```
wanliAlex, line 9: feel free to remove the mark if you believe all the tests are fixed. Also update mark logic in the conftest.py

tests/v2_tests/test_index_init_logging.py
```diff
@@ -40,51 +39,54 @@ def test_index_init_creating(self, mock_warning, mock_get):
...
+
+                assert mock_warning.call_count == 0
+
+                marqo_url_and_version_cache.clear()
+
+                # instantiating an index should trigger a warning, as the version is
+                # not supported
+                ix = temp_client.index(test_index_name)
+                mock_get.assert_called_once()
+                assert mock_warning.call_count == 1
```
wanliAlex, line 78: nit: use self.assertEqual(expected_value, returned_value) is better

wanliAlex (changes requested):
Ask pandu about the diversity of indexes issue

tests/cloud_test_logic/cloud_test_index.py
pandu-k: I think this is good for now. we should add more index types, now that the k8s will be more reliable

tests/marqo_test.py
pandu-k: we should. another index can work

pandu-k: thanks for catching this Li

tests/marqo_test.py
pandu-k: @wanliAlex the open source tests should be more thorough in testing open source features and should be "purer" as they don't have to work around the constraints of setting up infrastructure in a remote cloud environment. 

The open source version of the tests should test marqo/py-marqo features more rigorously. the cloud versions are there to ensure the cloud is working as expected and that it isn't reducing any Marqo functionality

tests/v2_tests/test_create_index.py
danyilq: Not all of them are resolved as some of them are bugs on marqo side like number of deleted docs is incorrect in response so I keep tag for now, we can remove it after fixing some of remaining tests in next marqo release that would fix some of the bugs that needs to be addressed.

wanliAlex (dismissed):

wanliAlex (approved):

PR #216, opened 2024-01-22
Danyil/fix cloud cleanup

danyilq:

* **Other information**:

### DIFF
--- a/tests/cloud_test_logic/cloud_test_index.py
+++ b/tests/cloud_test_logic/cloud_test_index.py
@@ -74,7 +74,10 @@ index_name_to_settings_mappings = {
             {"name": "int_filter_field_1", "type": "int", "features": ["filter", "score_modifier"]},
             {"name": "bool_field_1", "type": "bool", "features": ["filter"]},
         ],
-        "tensorFields": ["text_field_1", "text_field_2", "text_field_3", "image_field_1"]
+        "tensorFields": ["text_field_1", "text_field_2", "text_field_3", "image_field_1"],
+        "imagePreprocessing": {
+            "patchMethod": "simple",
+        }
     },
     # CloudTestIndex.unstructured_text_custom_prepro: {
     #     "type": "unstructured",
--- a/tests/cloud_test_logic/cloud_test_index.py
+++ b/tests/cloud_test_logic/cloud_test_index.py
@@ -32,7 +32,7 @@ class CloudTestIndex(str, Enum):
 
     unstructured_text = "pymarqo_unstr_txt"
     unstructured_image = "pymarqo_unstr_img"
-    unstructured_text_custom_prepro = "pymarqo_unstr_txt_custom_prepro"
+    unstructured_text_custom_prepro = "pymarqo_unstr_txt_cstm_pre"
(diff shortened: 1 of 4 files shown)


tests/cloud_test_logic/cloud_test_index.py
```diff
@@ -30,14 +30,14 @@ class CloudTestIndex(str, Enum):
         -> 4) use structured_image_prepro
     """
 
-    unstructured_text = "unstr_txt"
-    unstructured_image = "unstr_img"
-    unstructured_text_custom_prepro = "unstr_txt_custom_prepro"
+    unstructured_text = "pymarqo_unstr_txt"
+    unstructured_image = "pymarqo_unstr_img"
+    unstructured_text_custom_prepro = "pymarqo_unstr_txt_custom_prepro"
```
wanliAlex: Is this index created during the test? There is a restriction on the length of the index name (32 chars). Note we add 4 chars as unique identifiers in addition to this string

tests/cloud_test_logic/delete_all_cloud_test_indexes.py
```diff
@@ -13,18 +13,19 @@ def delete_all_test_indices(wait_for_readiness=False):
...
+    print(f"- index name starts with '{prefix}'")
     print(f"- index name contains the value of the environment variable MQ_TEST_RUN_IDENTIFIER: {suffix}\n")
     client = marqo.Client(**local_marqo_settings)
     indexes = client.get_indexes()
     indices_to_delete = []
     for index in indexes['results']:
-        if index["indexName"].startswith('test_index'):
-            if suffix is not None and suffix in index["indexName"].split('_'):
+        if index["indexName"].startswith(prefix):
+            if suffix is not None and index["indexName"].endswith(suffix):
```
wanliAlex, line 28: Will this also work with the appended unique identifier?

wanliAlex (changes requested):

tests/cloud_test_logic/delete_all_cloud_test_indexes.py
danyilq: unique identifier is the suffix it checks for.

tests/cloud_test_logic/cloud_test_index.py
danyilq: This one is actually not used anywhere and it just needs to be renamed.

wanliAlex: Can you make sure this index can be created correctly and the custom model can be used? Because this is related to custom models which is kind of important.

danyilq: I could add it and add another test but it will force us to have 4 indexes created per test execution. @pandu-k should we do it?

pandu-k: custom models are an important use case. Ideally we find a way to make it work with the current indexes, but if we need to add a new index for this to work we can do that too.

Tbh when we fix this, this test should be extended to include private (auth-required) models

wanliAlex (approved):

PR #231, opened 2024-05-08
added content_type fields to embed function and test requests

RaynorChavez:

Added the `content_type` parameter to the embed function and `content_type="document"` to embed test requests, which is a new field added in https://github.com/marqo-ai/marqo/pull/821 for the prefix feature

- Without adding the content_type as a parameter in the embed function, we are not able to pass its value into marqo.
- Without defining the field content_type, content_type defaults to "query" prefixes, which makes the tests fail when comparing to embeddings from add_documents since add_documents default to "document" prefixes.

- We add content_type as a parameter and are now able to take this field.
- By defining content_type defaults="document" we make sure that the prefixes prepended to document chunks are the same for both embed calls and add_document calls.

None

* **Other information**:

### DIFF
--- a/src/marqo/enums.py
+++ b/src/marqo/enums.py
@@ -25,3 +25,7 @@ class InterpolationMethod(str, Enum):
     LERP = "lerp"
     NLERP = "nlerp"
     SLERP = "slerp"
+
+class EmbedContentType(str, Enum):
+    Query = "query"
+    Document = "document"
--- a/src/marqo/enums.py
+++ b/src/marqo/enums.py
@@ -26,6 +26,7 @@ class InterpolationMethod(str, Enum):
     NLERP = "nlerp"
     SLERP = "slerp"
 
+
 class EmbedContentType(str, Enum):
     Query = "query"
     Document = "document"
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -376,7 +376,7 @@ class Index:
 
     def embed(self, content: Union[Union[str, Dict[str, float]], List[Union[str, Dict[str, float]]]],
               device: Optional[str] = None, image_download_headers: Optional[Dict] = None,
-              model_auth: Optional[dict] = None):
+              model_auth: Optional[dict] = None, content_type: Optional[str] = None):
         """Retrieve embeddings for content or list of content.
         Args:
             content: string, dictionary of weighted strings, or list of either. Strings
@@ -404,6 +404,7 @@ class Index:
         )
         body = {
             "content": content,
+            "content_type": content_type,
         }
 
         if image_download_headers is not None:
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -405,13 +405,14 @@ class Index:
         )
         body = {
             "content": content,
-            "content_type": content_type,
         }
 
         if image_download_headers is not None:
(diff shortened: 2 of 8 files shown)


src/marqo/index.py
```diff
@@ -404,6 +404,7 @@ def embed(self, content: Union[Union[str, Dict[str, float]], List[Union[str, Dic
         )
         body = {
             "content": content,
+            "content_type": content_type,
```
wanliAlex, line 416: Do you still want to include this in the body even if it is `None`?

src/marqo/index.py
```diff
@@ -376,7 +376,7 @@ def recommend(self, documents: Union[List[str], Dict[str, float]],
 
     def embed(self, content: Union[Union[str, Dict[str, float]], List[Union[str, Dict[str, float]]]],
               device: Optional[str] = None, image_download_headers: Optional[Dict] = None,
-              model_auth: Optional[dict] = None):
+              model_auth: Optional[dict] = None, content_type: Optional[str] = None):
```
wanliAlex: doc string for content type

src/marqo/index.py
```diff
@@ -376,7 +376,7 @@ def recommend(self, documents: Union[List[str], Dict[str, float]],
```
wanliAlex, line 384: If we only accept `document` and `query` for content type, can you provide an Enum for it?

wanliAlex (changes requested):

src/marqo/index.py
RaynorChavez: Resolved

src/marqo/index.py
RaynorChavez: provided `EmbedContentType`

src/marqo/index.py
RaynorChavez: Yeah I think it's alright. the embed endpoint in Marqo handles `None`, `query`, or `document`

src/marqo/index.py
wanliAlex: This should be Optional[your enum]

src/marqo/index.py
wanliAlex: I am not 100% sure about this point. I would say we should omit it because we are omitting `image_download_headers` and `model_auth`.  @farshidz Can you confirm this?

src/marqo/enums.py
```diff
@@ -25,3 +25,7 @@ class InterpolationMethod(str, Enum):
     LERP = "lerp"
     NLERP = "nlerp"
     SLERP = "slerp"
+
```
wanliAlex, line 28: nit: two lines between classes

wanliAlex (changes requested):

src/marqo/enums.py
RaynorChavez: Resolved

src/marqo/index.py
RaynorChavez: In the meantime, I made its inclusion conditional like `image_download_headers` and `model_auth`

RaynorChavez: Actually according to the expected cases from Joshua and Owen, we really do need to include content_type even if it is None,

farshidz (approved):

PR #265, opened 2024-11-22
Unpin urllib3 (remove upper bound) to allow 2.x version to be used

papa99do:

Dependency versioning

- urllib3 are pinned to 1.26.x version

- urllib3 are only pinned to >=1.26.0 version

No

* **Other information**:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,13 +8,11 @@ setup(
     install_requires=[
         # client:
         "requests",
-        "urllib3<2.0.0, >=1.26.0",
+        "urllib3>=1.26.0",
         "pydantic>=2.0.0",
-        "typing-extensions>=4.5.0",
         "packaging"
     ],
     tests_require=[
-        "pytest",
         "tox"
     ],
     name="marqo",
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         "tox"
     ],
     name="marqo",
-    version="3.9.1",
+    version="3.9.2",
     author="marqo org",
     author_email="org@marqo.io",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -8,13 +8,11 @@
     install_requires=[
         # client:
         "requests",
-        "urllib3<2.0.0, >=1.26.0",
+        "urllib3>=1.26.0",
         "pydantic>=2.0.0",
-        "typing-extensions>=4.5.0",
```
papa99do, line 13: this dep is only introduced to use @deprecated annotation. Since we do not need this dep now, I removed this restraint.

setup.py
```diff
@@ -8,13 +8,11 @@
...
         "packaging"
     ],
     tests_require=[
-        "pytest",
```
papa99do, line 17: pytest is installed by tox. Since we use tox to run tests, this section in setup.py is not used any more.

wanliAlex (approved):
LGTM

setup.py
```diff
@@ -8,13 +8,11 @@
...
         "tox"
     ],
     name="marqo",
```
wanliAlex, line 18: Do you want to upgrade the version here to publish it to Pypi?

wanliAlex (changes requested):

setup.py
papa99do: will do. @wanliAlex , this should be a new minor version, right?

wanliAlex: Yes, sounds like a patch version.

wanliAlex (approved):
LGTM

PR #269, opened 2025-01-06
Add rerank count parameter to search

vicilliar:

feature

no rerank_count parameter in search

added new parameter

no

* **Other information**:

### DIFF
--- a/src/marqo/index.py
+++ b/src/marqo/index.py
@@ -210,7 +210,8 @@ class Index:
             raise UnsupportedOperationError("This operation is only supported for Marqo Cloud")
 
     def search(self, q: Optional[Union[str, dict]] = None, searchable_attributes: Optional[List[str]] = None,
-               limit: int = 10, offset: int = 0, search_method: Union[SearchMethods.TENSOR, str] = SearchMethods.TENSOR,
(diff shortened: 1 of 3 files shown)


src/marqo/index.py
```diff
@@ -210,7 +210,8 @@ def get_status(self):
             raise UnsupportedOperationError("This operation is only supported for Marqo Cloud")
 
     def search(self, q: Optional[Union[str, dict]] = None, searchable_attributes: Optional[List[str]] = None,
-               limit: int = 10, offset: int = 0, search_method: Union[SearchMethods.TENSOR, str] = SearchMethods.TENSOR,
+               limit: int = 10, offset: int = 0, rerank_count: Optional[int] = None,
```
farshidz: Update minimum compatible version

farshidz (changes requested):

src/marqo/index.py
farshidz: This is actually a breaking change. You can only add arguments to the end of the list as users could be calling this based on the order only

farshidz (changes requested):

src/marqo/index.py
vicilliar: You're right, great catch, will move this to the end.

farshidz (approved):

PR #272, opened 2025-01-28
add index name header to all requests

vicilliar:

feature

http requests only have api key in header (if present)

http requests send index name in header

no

* **Other information**:

### DIFF
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -2,9 +2,7 @@ import copy
 import json
 from json.decoder import JSONDecodeError
 from typing import get_args, Any, Callable, Dict, Literal, List, Optional, Tuple, Union
-
 import requests
-
 from marqo.config import Config
 from marqo.errors import (
     MarqoWebError,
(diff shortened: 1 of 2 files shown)


vicilliar:
Cloud integration tests: https://github.com/marqo-ai/py-marqo/actions/runs/13004448466/job/36268777672

Only 2 tests failing are due to cloud staging not being on 2.15 yet. 
The other tests pass, meaning the addition of index name headers to all requests does not break existing functionality.

src/marqo/_httprequests.py
```diff
@@ -57,6 +56,7 @@ def send_request(
         index_name: str = ""
     ) -> Any:
         req_headers = copy.deepcopy(self.headers)
+        req_headers[CLOUD_INDEX_NAME_HTTP_HEADER] = index_name
```
VitusAcabado, line 59: What should we do if `index_name == ""`?
I'm pretty confident that it won't break anything but should we add this header in the first place if `index_name` argument isn't passed?

vicilliar: Actually I'd prefer if we don't send any index name header at all in this case. This will be for requests like `mq.get_indexes()`. Do those need the index name header in gcp?

VitusAcabado: It depends, if the client was instantiated with the marqo cloud api endpoint, it is not needed, however if it is instantiated with the index endpoint directly, then yes it is required.

vicilliar: Then in this case, it is better to always pass the header `x-marqo-index-name: ""` if no index name is provided, just so no request is missing the header.

wanliAlex (approved):
LGTM

VitusAcabado (approved):

PR #299, opened 2025-08-21
Add allow_missing_documents and allow_missing_embeddings to recommend

wanliAlex:

Add allow_missing_documents and allow_missing_embeddings to recommend

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         "tox"
     ],
     name="marqo",
-    version="3.16.0",
+    version="3.16.1",
     author="marqo org",
     author_email="org@marqo.io",
     description="Tensor search for humans",
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         "tox"
     ],
     name="marqo",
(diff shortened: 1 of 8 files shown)


src/marqo/version.py
```diff
@@ -1,4 +1,4 @@
-__minimum_supported_marqo_version__ = "2.22.0"
+__minimum_supported_marqo_version__ = "2.22.1"
```
papa99do: change this later to make the test pass first

setup.py
```diff
@@ -15,7 +15,7 @@
         "tox"
     ],
     name="marqo",
-    version="3.16.0",
+    version="3.16.1",
```
papa99do: change this later to make the test pass first

papa99do (commented):
So we don't need to add to search context? Is it because it's a dict?

src/marqo/version.py
papa99do: min version has to be 2.22.2. I'll create a release for 2.22.1 (just some cloud related changes).

papa99do (approved):

